PR #28, opened 2016-05-27
db migration support medic/medic-webapp#2323

SCdF:
**WIP DO NOT MERGE**

Removes the custom db setup code and replaces it with a library that
knows how to migrate between different versions. This also moves the
database setup from the couch2pg library into the main index file, the
logic there being that both couch2pg and xmlforms require that the db
is setup to be useful.

I also removed the periodic error logging in index in favour for just
one at the end, the reasoning being that if anything goes wrong we
should stop the promises chain at that point (logging errors "clears"
them and allows the chain to continue).


SCdF:
Superseded by https://github.com/medic/medic-analytics/pull/29

PR #30, opened 2016-06-07
medic/medic-webapp#2124 Reflect edited and deleted documents from couchdb

SCdF:
Adds DB migration support and teaches couch2pg to understand edits and deletes.


index.js
```diff
@@ -18,30 +19,48 @@ var startTime = function() {
   });
 };
 
+var potentiallyMigrateDatabase = function() {
```
btbonval: I feel a little weird about this function being defined directly in the integration piece of code, which (in my head) was meant to tie disparate pieces of code together. For OCD purposes, I might pull this function definition out and put it into its own file. However, I'm not sure OCD is a good reason to move things around ;)

index.js
```diff
@@ -18,30 +19,48 @@ var startTime = function() {
...
 var loop = function () {
   var starttime;
   startTime()
     .then(function (time) {
       starttime = time;
     })
-
+    .then(potentiallyMigrateDatabase)
+    .then(function(migrations) {
+      console.log('Sucessfully migrated', migrations);
```
btbonval: Missing a c: "Successfully."

index.js
```diff
@@ -18,30 +19,48 @@ var startTime = function() {
...
+
 var loop = function () {
   var starttime;
   startTime()
     .then(function (time) {
       starttime = time;
     })
-
+    .then(potentiallyMigrateDatabase)
+    .then(function(migrations) {
```
btbonval: I don't understand where the return value comes out of potentiallyMigrateDatabase.

I see a single resolve under the '201505271423' timestamp.

How does this work if there were more migrations (each with a resolve)? It is just a list that gets built from concatenating the result of thenables, with each migration as a following then?

index.js
```diff
@@ -18,30 +19,48 @@ var startTime = function() {
...
+      console.log('Sucessfully migrated', migrations);
+    })
     .then(couch2pg)
     .then(function () {
       console.log('Imported successfully at ' + Date());
-    }, function (err) {
-      console.log('Import errored at ' + Date());
```
btbonval: We might want to leave this in if there is no other location where an error is explicitly confirmed with a timestamp. There might be exactly that, in which case this is fine.

index.js
```diff
@@ -18,30 +19,48 @@ var startTime = function() {
...
     .then(function () {
       console.log('XML forms completed at ' + Date());
-    }, function (err) {
-      console.log('XML forms errored at ' + Date());
-      console.log(err);
     })
-
-    .then(function () {
+    .catch(function(err) {
+      console.error(err);
```
btbonval: Originally, errors thrown by couch2pg did not prevent xmlforms from running. It is true that xmlforms consumes couch2pg, but in practice, I have seen occasions where couch2pg produces more data than xmlforms can consume in an interval. In cases like that, it made sense to continue running xmlforms parsing even when couch2pg failed. Perhaps this no longer holds with the changes to couch2pg.

We should definitely throw a timestamp on here regardless.

btbonval:
The db migration stuff is great. In the back of my head, I thought it would be a nice idea, but I wasn't sure if it was necessary. Seeing it in place, and what it replaced, I think the migration work is a really good idea. 

Looking over the migration commit first, I added a few comments but generally I like it. The database handling is much cleaner.

index.js
btbonval: I think I see where the logging is handled instead and had commented there.

I'm going to delete this comment and the one previous.

btbonval:
I think you changed how the migration names work from dates to sequence numbers.

I have found that sequence numbers for migration names lend themselves to commit conflicts when people are working simultaneously. Dates, especially down the the second, for migration names tend to allow better parallel work.

It also allows for easier fixes to migrations that didn't go as planned. Migration bugs can be hard to catch in testing. When problems were found on staging or production system, I have seen cases where changing the last-added migration doesn't really work, and instead it is easier to add a new migration afterwards (this is not part of the argument) or add a new migration previous to the last-added migration (this is a bit harder to do when your migrations are a sequence).

There are probably pretty solid arguments against dates as migration names as well. Migrations can get hairy and there might be better ways to handle the problems I'm talking about above.

btbonval:
It would probably be useful to update the package number before we merge this so that there is no confusion that 0.0.3 is the before time, the long long ago, and this PR begins the new stuff.

.travis.yml
```diff
@@ -20,10 +19,11 @@ addons:
    postgresql: "9.4"
 
 before_script:
-   - psql -c "CREATE ROLE full_access;" -U postgres
+   # Used by (so far non-existent) spec tests for xmlforms
```
btbonval: I see references here to a postgresql database used for couch2pg integration tests:
https://github.com/medic/medic-analytics/blob/2124.reflect-edited-and-deleted-rebased/spec/couch2pg/index.js#L8-L12

How does that database get created in Travis?

README.md
```diff
@@ -13,103 +13,37 @@ are required.
...
 
 ## Required database setup
 
-It is assumed PostgreSQL is in use.
+We support PostgreSQL 9.4 and greater. The user passed in the postgres url needs to have full creation rights on the given database.
 
-### `full_access` role
+## Running tests
 
-A role called `full_access` must exist, and the role accessing PostgreSQL must
```
btbonval: A lot of the content removed here documents how we have setup two products currently in production and includes instructions that are not actually limited to couch2pg.

We should find somewhere to move these instructions for creating new production databases which mirror those two products. There will be one which follows the same pattern.

I'm not really sure where we should put this sort of IT setup documentation.

index.js
btbonval: nevermind. it was moved out in the next commit.

index.js
btbonval: Nevermind. Changed to something else in the next commit.

index.js
btbonval: Nevermind. In the next commit, It looks like there is no longer a migrations argument being returned from `potentiallyMigrateDatabase`.

index.js
btbonval: This changed slightly in the next commit, but I'm still not entirely sure if the error catching works the same as before, so I'll leave the comment here.

Similarly, I think we should add a timestamp onto the equivalent to this line in the next commit.

index.js
```diff
@@ -1,69 +1,37 @@
...
+      console.error('Something went wrong', err);
     });
 };
 
-loop();
-setInterval(loop, sleepMins);
+loop().then(function() {
+  console.log('Next run at ' + new Date(new Date().getTime() + sleepMs));
+  setInterval(loop, sleepMs);
+});
```
btbonval: If I'm reading this correctly, it resolves https://github.com/medic/medic-webapp/issues/2151

btbonval: Oh, I don't think I did read it correctly. It looks instead like you've made the code more consistent but effectively it is the same.

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+    return db.query(
+      format('DELETE FROM couchdb WHERE doc->>\'_id\' in (%L)',
+        docIdsToDelete));
+  } else {
+    return Promise.resolve();
+  }
+};
+
+var storeSeq = function(db, seq) {
+  return db.query(format('UPDATE couchdb_progress SET seq = %L', seq));
```
btbonval: I almost wonder if we could add a JSON object to the `couchdb` table with this same information, rather than create a whole new table for a single value. On the other hand, this gives us a place to grow and avoids possible quantum implosion events from storing the seqnum with all the stuff that gets fetched using the seqnum.

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+};
+
+/*
+NB: loadAndStore doesn't try to only load and store documents that have changed
+from postgres' perspective, because we presume if something appears in the
+changes feed it needs updating.
```
btbonval: This clarification helps reading the code below a lot. Thank you for the NB.

At first I thought everything was getting deleted and had to come back to this note.

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+    var changeSet = docsToDownload.splice(0, concurrentDocLimit);
+    var maxSeq = _.max(_.pluck(changeSet, 'seq'));
+
+    return couchdb.allDocs({
+      include_docs: true,
+      keys: _.pluck(changeSet, 'id')
+    }).then(function(couchDbResult) {
+      console.log('Pulled ' + couchDbResult.rows.length + ' results from couchdb');
+
+      console.log('Clearing any existing documents from postgresql');
```
btbonval: The phrasing on this led me to misinterpret it. Perhaps, to back the assumption we're making here, it would be good to explicitly state "updated docs" as opposed to "any existing docs". "Clearing any updated docs from postgresql" or something?

btbonval: Actually, if I read this right, it is also deleting "deleted docs". Say that 5 times fast.

"Clearing any updated or deleted docs from postgresql" or something?

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+changes feed it needs updating.
+
+If somehow this changes we can make this code more complicated.
+*/
+var loadAndStoreDocs = function(db, couchdb, concurrentDocLimit, docsToDownload) {
+  if (docsToDownload.length > 0) {
+    var changeSet = docsToDownload.splice(0, concurrentDocLimit);
+    var maxSeq = _.max(_.pluck(changeSet, 'seq'));
+
+    return couchdb.allDocs({
```
btbonval: I notice that `_rev` is not used. It sounds like changes feed sort of takes care of any rev considerations for you as when you supply the seqnum.

Is that correct?

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+      return deleteDocuments(db, _.pluck(couchDbResult.rows, 'id'))
+        .then(function() {
+          return couchDbResult;
+        });
+    }).then(function(couchDbResult) {
+      console.log('Inserting ' + couchDbResult.rows.length + ' results into couchdb');
+
+      return db.query(format(
+        'INSERT INTO couchdb (doc) VALUES %L',
+        couchDbResult.rows.map(function(row) {
```
btbonval: I'm confused here. So in the previous function, we deleted all the couchDbResult doc ids as they were either updated or deleted.

Here, without any obvious filter I can see to differentiate delete from update, we are adding back all of the same doc ids.

I guess I'm not clear on how a deleted doc does not get added back. Shouldn't deleted docs be filtered out somewhere so that this only adds back updated docs? I'm probably missing something obvious.

btbonval:
Quick note: I think this output might become very verbose. We definitely like to check on the adapter from time to time just to look at timestamps and make sure imports and xml parsing has been done recently and without overlapping itself.

It looks like `console.log` is being used for a lot of debug, but also informational waypoints like timestamps for each run. `console.error` is being employed for errors. Unfortunately `console.info` is an alias for `console.log` in Node. It would be nice if we could have debug statements be expressed conditionally when verbose or debug is enabled.

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+          return [JSON.stringify(row.doc)];
+        })));
+    }).then(function() {
+      return storeSeq(db, maxSeq);
+    }).then(function() {
+      console.log('Marked seq at ' + maxSeq);
+
+      return loadAndStoreDocs(db, couchdb, concurrentDocLimit, docsToDownload);
```
btbonval: This is interesting. So instead of running on a timer or putting a delay in, when `loadAndStoreDocs` completes its promise chain, the next `loadAndStoreDocs` begins processing _immediately_.

It looks like recursion, but that's just a trick of light and shadow caused by the sun's reflection on the thenable.

libs/couch2pg/importer.js
```diff
@@ -0,0 +1,105 @@
...
+          })
+          .then(function() {
+            console.log('Marked final seq of ' + changes.last_seq);
+            return storeSeq(db, changes.last_seq);
+          });
+      });
+  };
+
+  return {
+    import: _import
```
btbonval: Why define `function _import` only to return an object with property `import`? Is it to save tabs?

I would have probably done
```
return {
  import: function () {
    ...
  }
}
```

but I can see how that increases the tabs across the board.

libs/couch2pg/importer.js
btbonval: I see, this whole section does not execute when `docsToDownload` becomes empty, so `loadAndStoreDocs` stops calling itself when the `docsToDownload` has cleared.

libs/couch2pg/importer.js
btbonval: I read below where it appears deleted documents are handled separately from edited documents. `loadAndStoreDocs` strictly handles edited docs.

libs/couch2pg/index.js
```diff
@@ -1,51 +1,36 @@
...
+    postgrator.setConfig({
+      migrationDirectory: __dirname + '/migrations',
+      schemaTable: 'couch2pg_migrations',
+      driver: 'pg',
+      connectionString: POSTGRESQL_URL
+    });
 
-  // establish a single connection to postgresql
-  return pglib({ 'promiseLib': Promise })(process.env.POSTGRESQL_URL)
-    .connect()
```
btbonval: i don't see `.connect()` or any task handling from pg-promise.

This means every `db.query` in `importer.js` is going to create a new connection. I count 2 DB connections for seq and the first delete, then 2*N DB connections as `loadAndStoreDocs` slams through N pages of docs (each page being no more than concurrentDocLimit).

Each iteration over `process.env.COUCH2PG_SLEEP_MINS` would open 2+2*N postgres DB connections.

This might be acceptable as long as N does not get large for too long. I was trying to keep each iteration sharing a single DB connection, but that is not done by default with pg-promise queries.

libs/couch2pg/migrations/001.do.2318-standardise-schema.sql
```diff
@@ -0,0 +1,19 @@
+-- get everyone to baseline what existed before we added migrations
+CREATE TABLE IF NOT EXISTS couchdb (doc jsonb);
+
+-- IF NOT EXISTS doesn't exist (!) in Postgres 9.4, so do this silliness instead
+DO $$
+BEGIN
+
+IF NOT EXISTS (SELECT 1 FROM pg_class c WHERE c.relname = 'couchdb_doc_uuid') THEN
```
btbonval: We could probably abstract these three lines into a PG Function called `IF_INDEX_NOT_EXISTS()`, which would then be available to these three index commands.

I think that might actually be the way to go in the future because we might need to make some dynamic indices based on discovered form names if we keep XML parsing separate. 

https://github.com/medic/medic-webapp/issues/2232

libs/xmlforms/index.js
```diff
@@ -1,82 +1,115 @@
...
-      }
-      return formreport.storeReports(db, pgsql, reports);
-    }, handleError)
-    .then(function () {
-      if (process.env.COUCH2PG_DEBUG) {
-        console.log('refreshing materialized views');
+
+      if (_.isEmpty(reports)) {
+        return new Promise(function(succ) { succ(); });
+      } else {
```
btbonval: This level is nested logic is usually something I'd toss into its own file, leaving the integration code to be a collection of functions defined elsewhere.

I think, as written, it would be very difficult to unit test all of the indented logic under this.

SCdF:
> I think you changed how the migration names work from dates to sequence numbers.

> I have found that sequence numbers for migration names lend themselves to commit conflicts when people are working simultaneously. Dates, especially down the the second, for migration names tend to allow better parallel work.

You what, I liked them too! Unfortunately the migration tool used ints and not longs in its migration table (so an timestamp is an int too large). I considered raising a bug with PR, forking and using that until he fixed his version, but I decided that the benefits weren't worth the time.

But since there is another vote for it I'll give it a crack today. It should really only take me 10 minutes.

.travis.yml
SCdF: It gets created in the `initialistPosgres` function which gets called before any tests run: https://github.com/medic/medic-analytics/blob/2124.reflect-edited-and-deleted-rebased/spec/couch2pg/index.js#L58-69

I don't really want to do any scripting in the travis file if I can avoid it, because the travis file doesn't get used locally, and you should be able to run all tests locally.

README.md
SCdF: Yeah me neither. Apart from "in the code" I don't know about any Medic documentation. I presume this is something a Tech Lead would know about? Regardless, I agree with you that it should be somewhere in our internal documentation and not our open source space.

index.js
SCdF: Yep it should resolve medic/medic-webapp#2151 :-) The code inside of `then` should only run once all the code in loop is complete.

libs/couch2pg/importer.js
SCdF: There is definitely a future where you'd store a JSON object for configuration / meta information, which would occur if we started storing more data. For example, I've considered storing the couchdb and pg db you connect to and having them added in a setup stage (which is different from the run stage), because right now I don't like that the sequence is stored away from and incidentally to the couchdb and pg urls, even though they are indelibly connected. I will probably raise a ticket for us to move to that at some point, but I didn't want to dig too deep right now.

However, I wouldn't store them in the couchdb table. I'd want it to be clear that the couchdb table is for data that is pulled from couchdb, and if you see a piece of data in that table, you know that it is currently (relative to the last sync) in couchdb in that exact state.

libs/couch2pg/importer.js
SCdF: Yeah so the somewhat confusing "any existing documents" phrase is to cover the fact that you have a selection of ids, some may be completely new and some may be edits to existing documents. So you're passing a bunch of ids to delete, some of which may not already exist in PG.

Happy to take suggestions for better phrasing :-)

libs/couch2pg/importer.js
SCdF: The changes feed gives you a lot of stuff, because you can do crazy things with this algorithm / protocol, and your database (if this were couch->couch) could be in lots of different states.

For us though, we only "replicate" from one couchdb, and so we know that if there is an entry in that changes feeds we care about the document, and we care about the latest version.

Note that we're not doing anything clever like dealing with conflicts. Neither does medic-webapp either, so it's OK for now.

libs/couch2pg/importer.js
SCdF: It sounds like in other comments on this PR you sorted out your understanding of this. Let me know if that's not the case though and I'll explain :-)

SCdF:
> Quick note: I think this output might become very verbose.

Oh yeah no doubt. It still suffers from being written so I could see what it was doing (until I discovered you can debug node in chrome). I'll make a note to hide most of it behind a debug flag.

As to whether a debug flag is the best idea: it's certainly the simplest idea. I presume there are logging libraries we could get from npm, but whatever, this is good enough for now.

libs/couch2pg/importer.js
SCdF: Yep! I mean… it's _sort of_ recursion. It's functionally the same thing, it is just expressed in this idiom a bit differently.

libs/couch2pg/importer.js
SCdF: Yeah it could go either way honestly. I just wanted all the funcs to be beside each other at the same level.

libs/couch2pg/index.js
SCdF: Good catch! So from reading the docs: "i.e. only the actual query methods acquire and release the connection.". I read that to mean what I am doing right now is inefficient (a new connection per action) but I'm not doing anything dangerous like leaking connections correct?

I'll make a note to change the code to open a new connection, pass it to `importer.js` and then close it once it's done though, because you're right, that is certainly better.

libs/couch2pg/migrations/001.do.2318-standardise-schema.sql
SCdF: Or we could move to PG9.5 :-)

I agree though, if we start doing this a lot we can add it as a permanent function into a migration script and use it elsewhere.

libs/xmlforms/index.js
SCdF: Probably! The change here was just the most minimal change to fix a bug where there were no new reports. FWIW the code flow is the same as it was before, I'm just wrapping what was there in an `_.isEmpty` check. Note also I haven't really looked at changing xmlforms much yet, that's what I'm doing presently.

index.js
SCdF: I'll add a timestamp to the error logging.

libs/couch2pg/index.js
SCdF: Hey @btbonval so I've read the docs a little more, and I'm actually OK with the current approach, as pg-promise (which wraps node-pg) uses connection pooling be default.

SCdF:
Hey @btbonval I've made some changes based on feedback. I also rebased and force-pushed because I missed some issue references in some earlier commits.

Actually new changes are these ones above:
image

btbonval:
> You what, I liked them too! Unfortunately the migration tool used ints and not longs in its migration table (so an timestamp is an int too large). I considered raising a bug with PR, forking and using that until he fixed his version, but I decided that the benefits weren't worth the time.

> But since there is another vote for it I'll give it a crack today. It should really only take me 10 minutes.

Ha! I actually deleted that comment because there were already too many and I thought it was ultimately quite superficial. I hadn't thought of a technical problem upstream, but yeah if the migration software cannot support a fairly common framework of YYYYMMDDhhmmss, that probably is a bug. Thanks for looking into it!

.travis.yml
btbonval: Fair point about local/Travis differentiation. I'll look more closely into where `initialisePostgresql()` is called, somehow I missed that function when tracing code yesterday, so I probably missed other things as well.

README.md
btbonval: Actually, I just recalled that Michael (being the largest user of postgres besides me at present) has a folder or two on our Drive for postgres related things. I prefer to version control things for no good reason, but I'll probably move it there. I think we're good on this thread.

libs/couch2pg/importer.js
btbonval: To check my understanding:

So, in a possible future, we might mark certain CouchDB docs with an id like `_dbname/pgconfig`, then the adapter could import that specifically from CouchDB and place it into a `couchdb`-like table containing other JSON docs specifically for adapter/postgres config. The same doc would probably end up in the `couchdb` table as a matter of course, unless we put filters on, but it wouldn't get used from that location.

If I got it, yeah that sounds like a good future plan.

btbonval: err I suppose there is the possibility of a separate couchdb database as well, and then such config could be the db's design doc.

libs/couch2pg/importer.js
btbonval: I think readability is more important. I honestly thought there was some kind of deference or function pointer cost, If this is a pretty common practice, then any such cost is probably trivial or nonexistent.

libs/couch2pg/index.js
btbonval: Nothing dangerous, no. It was more about postgres connection stability (and not getting locked out from too many connections). I will have to read more about the connection pooling as it applies to `node-pg` and as it filters up to `pg-promise`. I was led to believe from `pg-promise` docs that they don't use it.

libs/xmlforms/index.js
btbonval: Yeah after I made this comment, I sort of figured this whole xmlforms stuff might get ripped up anyway. ;)

btbonval:
> I also rebased and force-pushed because I missed some issue references in some earlier commits.

Sometimes I will create a second branch instead of force pushing, and then the previous branch will become an orphan that eventually gets `git branch -D` deleted (capital D necessary). It can make branch lists a mess though. :/

libs/couch2pg/importer.js
SCdF: So that's not what I was talking about, though it's an interesting idea.

What I meant was two things:
 - it makes sense to store the configuration for a postgresdb attaching to a couchdb in that postgresdb to keep that relationship clear and immutable (ie makes it harder to accidentally point a db at the wrong couchdb in the future or whatever)
 - storing configuration in RDBMS' is a bit of a weird one, and tends to get ungainly if every time you need a new configuration variable you add a new column to some table. IMO if you think you're going to store more than 2-3 things you might as well just store a JSON blob (or XML / YAML / EDN whatever) and parse it out.

garethbowen:
LGTM

PR #31, opened 2016-06-17
xmlforms deals with edited and deleted documents medic/medic-webapp#2124

SCdF:
couch2pg now returns a summary of changed documents. xmlforms then uses that
as a list of xmldata to delete. This works because new documents won't have
anything to delete, edited documents will be recreated and deleted docs will
actually be deleted.


index.js
```diff
@@ -21,16 +21,22 @@ var loop = function () {
...
-    })
-    .then(xmlforms.migrate)
-    .then(function() {
-      log.info('xmlforms Migration checks complete');
-    })
-    .then(xmlforms.extract)
-    .then(function () {
-      log.info('XML forms completed at ' + Date());
+
+      if (summary) {
```
garethbowen: I feel like there should always be a summary even if it's `{ edited: [], deleted: [] }`.

libs/xmlforms/cleaner.js
```diff
@@ -0,0 +1,60 @@
...
+
+    var grouped =_.groupBy(results, function(row) {
+      return 'formview_' + row.formname + '_' + row.formversion;
+    });
+
+    _.chain(grouped)
+      .keys()
+      .each(function(key) {
+        grouped[key] = _.pluck(grouped[key], 'uuid');
+      });
```
garethbowen: It'd be nicer to just use native JS here, eg: `Object.keys(grouped).forEach(function(key) {...`

libs/xmlforms/cleaner.js
```diff
@@ -0,0 +1,60 @@
...
+  return _.reduce(
+    _.keys(tableToDocs),
+    function(thenChain, key) {
+      return db.query(format('DELETE FROM %I WHERE xmlforms_uuid IN (%L)', key, tableToDocs[key]));
+    },
+    rsvp.Promise.resolve()
+    ).then(function() {
+      return tableToDocs;
+    });
+};
```
garethbowen: Isn't this really complicated? I think it'd be much clearer to map to an array of promises then use Promise.all.

libs/xmlforms/cleaner.js
```diff
@@ -0,0 +1,60 @@
...
+
+var deleteMetadata = function(db, tableToDocs) {
+  var allDocs = _.flatten(_.values(tableToDocs));
+
+  if (allDocs.length > 0) {
+    return db.query(format('DELETE FROM form_metadata WHERE uuid in (%L)', allDocs));
+  } else {
+    return rsvp.Promise.resolve();
```
garethbowen: IIRC if this is called within a `then` block you don't need to return a Promise.resolve() - if you `return` it'll just work.

garethbowen:
Looks good. Some style comments.

libs/xmlforms/cleaner.js
SCdF: It might be non-idiomatic for sure. Reading it, it's also wrong!: it should be `return then(function() { return db.query...`. 

The idea was to make them execute in serial so that we wouldn't have to deal with situations where we're occasionally blowing out postgres connection limits. Using LG's data there are already ~45 tables that we could be deleting from.

index.js
SCdF: Cool, will change

libs/xmlforms/cleaner.js
SCdF: Cool, will change

libs/xmlforms/cleaner.js
SCdF: Yeah that just clicked for me 5 minutes ago, once you have one promise the rest don't have to be.

PR #32, opened 2016-06-20
xmlforms learns about migrations medic/medic-webapp#2323

SCdF:
Xmlforms now uses a migrationt tool to setup its DB. Note that we now add
contacts tables etc even if you're 0.4, because why not: they'll just be
empty tables, nothing breaks.

I've also updated our postgrator version to the one with my fix in it
so we no longer rely on my fork.


garethbowen:
LGTM

PR #33, opened 2016-06-28
Remove XML parsing medic/medic-webapp#2432

SCdF:
Removes all XML parsing. Because there is no XML parsing we no longer
need to pass the changes from couch2pg (and couch2pg no longer needs
to generate them, but we do it anyway for now). Because we no longer
need to pass changes we don't need to worry about losing data, so we
run importAll() now instead.

All xmlforms does now is make sure the helper SQL exists (views and
the like) and that at the end of each run any materialised views that
exist are updated.


env.js
```diff
@@ -1,13 +1,14 @@
 var log = require('loglevel');
 
 module.exports = function() {
-  if (process.env.COUCH2PG_DEBUG) {
+  if (process.env.COUCH2PG_DEBUG || true) {
```
garethbowen: Accidental commit?

env.js
```diff
@@ -1,13 +1,14 @@
...
     log.setDefaultLevel('debug');
   } else {
     log.setDefaultLevel('info');
   }
 
   return {
+    legacyMode: process.env.LEGACY_MODE || false,
```
garethbowen: I'm not a fan of the name `legacyMode`. It'll become difficult once 2.6 is considered legacy. Can we give it a name that represents the feature we're interested in? ie: What's the real difference between legacy mode and non-legacy mode? For bonus points, can we detect this in the db rather than having to provide it?

index.js
```diff
@@ -11,50 +11,101 @@ var couch2pg = require('./libs/couch2pg/importer')(
...
+};
+var migrateXmlforms = function() {
+  return xmlformsMigrator(env.postgresqlUrl)();
+};
+
+var delayLoop = function(loopFn, errored) {
+  return new Promise(function(resolve) {
+    var ms = sleepMs(errored);
+    log.info('Run '+(errored ? 'errored' : 'complete') + '. Next run at ' + new Date(new Date().getTime() + ms));
+    if (ms === 0) {
```
garethbowen: This is confusing. I think it would be cleaner to say...

```
if (errored) { /* calculate ms and set interval */ }
else { resolve(); }
```

index.js
```diff
@@ -11,50 +11,101 @@ var couch2pg = require('./libs/couch2pg/importer')(
...
     } else {
-      return completeBatch(changes.lastSeq);
+      setInterval(resolve, ms);
     }
   });
 };
 
 var run = function() {
-  log.info('Beginning couch2pg and xmlforms run');
+  log.info('Beginning couch2pg and xmlforms run at ' + new Date());
```
garethbowen: Maybe we should find a logging library that outputs the timestamp...

index.js
```diff
@@ -11,50 +11,101 @@ var couch2pg = require('./libs/couch2pg/importer')(
...
+  })
+  .then(
+    function() {
+      return delayLoop(run);
+    },
+    function(err) {
+      log.error('XMLForms support failed');
+      log.error(err.stack);
+      return delayLoop(run, true);
+    })
```
garethbowen: Usually we use the `.catch` form, rather than passing in two functions to the `then`. Any reason you chose this form this time?

env.js
SCdF: So the intention here is that 0.4 is "legacy", and everything else, 2.6 and onward, is considered current. Thinking about it in the fresh light of day, perhaps this doesn't fit with Medic's deployment practices particularly well.

Currently the only difference is that if you have legacy mode turned on certain database migrations to add views get added, as those views are not relevant to 0.4.

I'll have a think about ways to make this clearly / feature toggled. In terms of detecting it we could possibly intentionally download the ddoc and get the version from that. Not sure we need that much structure yet though.

env.js
SCdF: Nope, intentional. The debug output is useful, and when this is properly deployed it should go into a log file.

At some point if we cbf we can replace this env var with an actual minimum level value that we default to debug or something.

index.js
SCdF: Yes. Loglevel (the library we're using) does this with a plugin, I just haven't got around to doing that yet. I should raise a ticket.

index.js
SCdF: Hmm, your code wouldn't do the same thing, so this is definitely too confusing!

Basically this is what's happening:
 - Calculate how long to wait (`sleepMs()`):
   - If there was no error it's whatever the sleep time is
   - If there was an error sleep for a amount of time based on a falloff period, so that if there is an error it retries immediately and then waits longer and longer with subsequent errors

Because both erroring loops and successful loops can wait, we need to calculate the ms in both instances. 

Let me know if that makes sense, and how we change the code to make that clearer.

index.js
SCdF: The reason was that it made it easier to be clearer where that boolean passed to `delayLoop` came from. Because JS is truthy I was worried that if that boolean was passed in   some refactor down the line would break it. One option was to change the boolean to something more specific, or to make sure everyone using the boolean did `if(errored === true)` instead of just `if(errored)` or whatever. I thought the more contained way was to make it an either or proposition and have the boolean explicit.

```
.then(function() {
  return doSomeStuff(); // [1]
}).catch(function(err) {
  log.error('XMLForms support failed');
  log.error(err.stack);
  return true;
})
.then(function(errored) {
  return delayLoop(run, errored);
});
```
In this example `[1]` either *has* to return nothing something falsey, or we have to add this in:

```
.then(function() { return false; })
```

So IDK, I wrote it the way I thought was the cleanest. It could definitely be the victim to a misunderstood refactor in the future though. I should either add a comment explaining why I'm using this style, or we should rewrite it to work differently. What are your thoughts? What makes the most sense to you?

env.js
garethbowen: Yeah Bryan and I talked about this at length. I wasn't really comfortable with version detection if there was a viable feature detection alternative. I can definitely see a 3.0 feature that adds views that aren't relevant to 2.6 so anything that's future proof will do.

env.js
garethbowen: Maybe just remove the dead code then?

index.js
garethbowen: Ok I see what you're saying now. I think on my first read through I misread one line, so ignore that comment.

The good news however is reading through it again I found another issue... the loopFn is never used. Maybe this is on purpose, but it still makes it confusing. My assumption was that the loopFn would be retried until it passed, but that's not what this does currently, right?

Secondly, I think you should use setTimeout not setInterval - the latter will attempt to resolve the promise multiple times.

It'd be good to chat through this line by line to make sure I'm understanding what's meant to be going on.

env.js
SCdF: Not sure which bit is dead exactly? You mean just default `loglevel` to output debug mode and if you want to change that change the code?

env.js
SCdF: I think doing this in a forward thinking way is really hard.

The problem with feature detection is that you never know if you have the whole story.

For example, we want some views to be created in 2.6 to make dealing with contacts easier. The detection, creation and management etc of this is now done using a DB migration tool. To do this with feature detection, we would need to build a system that checks for un-created views (if it creates two views, do you check one of them or both? What if you want to update that view or package of changes later?) and then determines if they need to be created: in this case, if there are any documents of type `contact`. You need to do this every time, because maybe the first time you sync up there are no contact documents yet. This seems really terrible compared to using a migration management tool.

Instead of feature detection you could group functionality broader than a single feature (allowing you to use things like db migration without it being a giant hassle), e.g. "xmlforms", which encompasses all work needed for that very broad feature. For example, in the future it may be required for us to extract out the form version from the XML and put that in a queryable table. Or extract out GPS coords from older forms that don't have them exposed in the JSON.

This doesn't really map well to a "feature" though, even a broad one like "xml forms". In future versions we could keep that broad feature but the code could change to not require (or even invalidate) the work that code needs to do. For example, in 3.0 we might already extract out the form version into the JSON, or maybe even have a system where the concept of form version doesn't exist anymore. 

As you said, you're not a fan of version detection, which is the next obvious step in this thought experiment, so we can skip over that one.

The route I was going with for "legacy mode" was that legacy == 0.4, not legacy == 2.x, *but only because the 1.0 of this comes out when both exist in production*. If we had built this when just 0.4 existed there would have been no legacy mode. When medic-webapp 3.0 comes along and is substantially different (or 2.8, or anything really), the idea was to simply not support 2.x in master anymore. The old releases still exist, and if there are bug fixes etc that are required we can check out the tag and go through that process. Once this changed version is a bit stable the idea would *also* be to drop legacy mode from master completely, and thus master of medic-couch2pg === master of medic-webapp, and everything else is tags.

garethbowen: We could at least call "legacyMode" something more descriptive like "0x-mode" (can't work out the right cases or whatever).

The whole thing about shipping versions together sounds good to me, but it does dictate our deployment infrastructure a little, because then we need to ensure they're upgraded together, which lends itself towards bundling as a gardener module IMO.

SCdF: We do have a ticket to put this in gardener, but while we're talking, I'm not really sure why? I mean, tags right? All I'm saying is that every release of couch2pg doesn't have to support every release of medic-webapp. As long as a) we tend toward upgrading people to the latest version of medic-webapp and b) couch2pg is a relatively simple piece of software that doesn't need updates much I don't see the need to get anything more complicated than that.

PR #35, opened 2017-04-24
Make medic-couch2pg die in case of CouchDB authentication failure

derickl:
Currently CouchDB authentication failures do not trigger an exception and this can go unnoticed for a long time. This change specifically tracks responses from CouchDB with a 401 status code and throws an exception so that we can handle the process exit via upstart.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ var couch2pg = require('couch2pg'),
     xmlforms = require('./libs/xmlforms/updater')(db);
 
 var firstRun = false;
+var authError = false;
 
 var errorCount = 0;
 var sleepMs = function(errored) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -59,20 +62,24 @@ var run = function() {
...
   var runErrored = false;
-
   return importer.importAll()
   .catch(function(err) {
     log.error('Couch2PG import failed');
     log.error(err);
 
+    // CouchDB will return a json response with a 401 status code in case of
+    // authentication failure
+    if ((typeof err == "object" && err.status == 401)){
```
SCdF: You don't need multiple parens here, you don't need to check the type and you do want to use triple equals instead of double equals. I should probably add a jshint step to the travis build so it catches stuff like that.

Anyway, one suggested replacement would be: `if (err && err.status === 401)`.

You should also drop the comment, this code self-documents (a 401 is a known status code meaning unauthorised and the block sets a var called `authError`).

index.js
```diff
@@ -18,6 +18,7 @@ var couch2pg = require('couch2pg'),
     xmlforms = require('./libs/xmlforms/updater')(db);
 
 var firstRun = false;
+var authError = false;
```
SCdF: this var does not prevent `firstRun` from being set to true, which while it does not matter in its current state, is not technically correct and could cause new issues in the code as things are refactored.

PR #36, opened 2017-11-09
Support for 2.13.x builds

SCdF:


SCdF:
For medic/medic-webapp#2635

PR #37, opened 2018-04-11
Updating to support medic and medic-sentinel replication

vimemo:
medic/medic-webapp#3423

@SCdF, here are some initial updates to support medic and medic-sentinel replication. Still would like to do more local testing but want to get your suggestions now.


index.js
```diff
@@ -58,7 +75,7 @@ var run = function() {
   log.info('Beginning couch2pg and xmlforms run at ' + new Date());
 
   var runErrored = false;
-  return importer.importAll()
+  return Promise.all[importer.importAll(), sentinel.importAll()]
```
SCdF: You need to change this to `sentinelImporter` :-)

vimemo:
@SCdF , please review. I have added some initial tests but you might have more ideas on more things to test. Things can be tested now with: 
docker-compose run test grunt test

.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules
 medic-analytics.sublime-project
 medic-analytics.sublime-workspace
+coverage
```
SCdF: What coverage tooling did we add?

.gitignore
```diff
@@ -1,3 +1,5 @@
...
+*.log
```
SCdF: What does this relate to?

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:carbon
```
SCdF: We need to control the version of node we're running.

If we're going with async await it should be 8.x, otherwise it should be 6.x

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: '2'
```
SCdF: You should update the README of this project to detail using this to run it locally, or run tests, as well as how to run it without this (for current prod deployments / people who hate docker, which on MacOS is a valid emotion to have :P)

libs/delay.js
```diff
@@ -0,0 +1,30 @@
+const log = require('./log')
+    , Promise = require('rsvp').Promise;
```
SCdF: I think we can drop rsvp for promises now! 😀

libs/delay.js
```diff
@@ -0,0 +1,30 @@
...
+
+let errorCount = 0;
+
+const delayLoop = (errored, sleepMs) => {
+  return new Promise(resolve => {
+    const ms = sleepTime(errored, sleepMs);
+    log.info('Run '+(errored ? 'errored' : 'complete') + '. Next run at ' + new Date(new Date().getTime() + ms));
```
SCdF: We're moving to ES6, so replace this with templated strings, e.g.:

PR #38, opened 2018-05-03
3423 incremental updates to support multiple db replications

vimemo:
@SCdF, here is the latest. I'll try to be available when you come back to chat about this. Updated things to avoid so many env variables and using the cli with some defaults instead. We can chat about it later to see how you feel about this :)


libs/medic.js
```diff
@@ -0,0 +1,135 @@
...
+      pgp = require('pg-promise'),
+      Promise = require('es6-promise').Promise,
+      log = require('./log');
+
+// Removes credentials from couchdb url
+// Converts http://admin:pass@localhost:5984/couch1
+// to localhost:5984/couch1 -- seq source
+const parseSource = url => {
+  const source = urlParser.parse(url);
+  return `${source.host}${source.path}`;
```
SCdF: Nice idea!

spec/bootstrap.js
```diff
@@ -0,0 +1,69 @@
...
+const waitForCouch = async (url) => {
+  await waitForDb({url: url, fn: async () => {
+    try {
+      const {statusCode} = await curl.get(url);
+      if(statusCode === 200) {
+        console.log(`- Couch [${url}] is now avaliable.`);
+        return true;
+      }
+    } catch(err) {
+      //Ignore
```
SCdF: Why would you want to ignore this?

spec/replicate.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('replicates', async () => {
+    expect(medicDocs.length + sentinelDocs.length).to.equal(4);
+
+    await medic(COUCH_DB_URL).replicateTo(PG_DB_URL, 1);
+    const pgconn = knex({client: 'pg', connection: PG_DB_URL});
+    let rows = await pgconn.raw('select * from couchdb');
+    expect(rows.rowCount).to.equal(4);//Avoids view
+    await pgconn.destroy();
```
SCdF: It doesn't really matter when there is one test, but perhaps this should be in an `afterEach`?

libs/runner.js
```diff
@@ -0,0 +1,68 @@
...
+
+const run = async (couchUrl, pgconn, timesToRun=undefined) => {
+  log.info('Beginning couch2pg and xmlforms run at ' + new Date());
+  const [allResults, runErrored] = await replicateAll(couchUrl, pgconn);
+  if(allResults) {
+    const [medicResults, sentinelResults] = allResults;
+    // Run secondary tasks if we reasonably think there might be new data
+    // runErrored || errorCount <- something went wrong, but maybe there is still new data
+    // firstRun <- this is first run, we don't know what the DB state is in
+    // results.deleted.length || results.edited.length <- data has changed
```
SCdF: This comment is now out of date. If you believe it's self-explanatory delete the comment, or if you think it's a useful comment keep it up to date.

spec/migrate.js
```diff
@@ -0,0 +1,40 @@
+const assert = require('chai').assert,
```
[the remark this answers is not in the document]
vimemo: This repo did not have any tests but was initially setup with a test directory for unit tests and spec for integration tests. I'll update to use just test dir.

spec/utils/pgutils.js
```diff
@@ -0,0 +1,39 @@
+const urlParser = require('url'),
+      knex  = require('knex');
+
+const ensureDbExists = async (url) => {
+  const opts = urlParser.parse(url);
+  const dbname = opts.path.slice(1);
+  const conn = knex({client: 'pg', connection: url.replace(opts.path, '')});
+  try {
+    await conn.raw(`CREATE DATABASE \"${dbname}\"`);
```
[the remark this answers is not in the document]
vimemo: Seems to be necessary for names like medic-analytics; dash.

spec/bootstrap.js
vimemo: good catch; pun intended. need to document or remove this.

package.json
```diff
@@ -1,35 +1,44 @@
 {
   "name": "medic-couch2pg",
-  "version": "1.2.1",
+  "version": "1.3",
```
SCdF: 2.0 plz, you've removed stuff, changed the way the CLI works etc, it definitely is an entirely new deal.

PR #39, opened 2018-12-20
Secure logs

vimemo:
Removes user/pass from db urls


libs/medic/safe.js
```diff
@@ -0,0 +1,18 @@
+const EXPECTED_URLS = ['couchdbUrl', 'postgresqlUrl'];
+
+const cleanUrl = (url) => {
+  if(url && url.split('@').length>0) {
+    return url.split('@').pop();
```
garethbowen: Don't split twice - store the output of the first split, eg:

```js
const parts = url && url.split('@');
if (parts && parts.length > 0) {
  ...
```

libs/medic/safe.js
```diff
@@ -0,0 +1,18 @@
...
+  }
+  return url;
+};
+
+const safe = (opts) => {
+  const o = Object.assign({}, opts);
```
garethbowen: Instead of `o` give it a descriptive name. What about `cleaned` or something?

vimemo:
@garethbowen, updated with your great review. thank you.

PR #44, opened 2019-02-08
Updates setup to test against different node versions 

vimemo:
Updates test image to build and run tests against different node versions.


vimemo:
@SCdF, good catch about tests only running against node 8. Here are updates to run tests against that travis matrix. thanks.

PR #46, opened 2019-03-15
Fix dependencies

SCdF:
Move to only importing those pieces of PouchDB we need

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "git://github.com/medic/medic-couch2pg.git"
   },
+  "bin": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,15 +7,19 @@
     "type": "git",
     "url": "git://github.com/medic/medic-couch2pg.git"
   },
+  "bin": {
+    "replicate-medic": "./index.js"
```
SCdF: @henokgetachew I also added this so you can boot it the same way as 3.x

henokgetachew: Yeah this is pretty cool.

Issue #54, opened 2019-12-05
Add ability to replicate medic-users-meta database to PostgreSQL


michaelkohn:
Now that we are storing telemetry data and feedback docs in `medic-users-meta`, we should make this data more accessible by either automatically replicating this database, or adding an environment variable that allows users to specify whether or not they want to replicate it (DEFAULT value if not provided could be NO). 

Instead of replicating this data to the `couchdb` table in PostgreSQL, we might consider replicating it to a separate table such as`couchdb_users_meta`.  While I don’t think it happens as much now, we’ve seen projects with 10s of thousands of feedback docs and it would be better to keep that data out of the main transaction table (`couchdb`).

yrimal:
Trying to make requirement more clear here:

1. Replicate `medic-users-meta` to the `couchdb_users_meta` table on same Postgres database where `medic`  is replicated. 
2. Since we don't have enough projects in latest CHT core where `medic-users-meta` exists, it's unclear whether `read:message:` docs are merged from `medic-user-{username}-meta` to `medic-users-meta` or not.  Neither App Services, Nor R&L team currently need `read:message` at the moment. So, we don't need to sync read report to postgres that will save us time and space as well. 

CC: @kennsippell

helizabetholsen:
Thanks for clarifying, @yrimal. This issue is definitely a priority for us to enable our further exploration of telemetry data, particularly as we look for cross-cutting impact metrics for both MM hosted and self-hosted partners.

MaxDiz:
added to v3.11 based on roadmap planning decision

yrimal:
I'd like to refrain from 

> 3. Ability to sync without using separate couch2pg process would be great.

And request that we make a separate process for syncing telemetry data. This thought comes considering that we might have partners coming in future who are self-hosted and are not willing to share real data or metrics but are open to share telemetry documents. In this situation, we'll be able to directly sync telemetry data from partner server to our rdbms monthly or in any specified period. 

Thoughts : @helizabetholsen , @yembrick  ?

michaelkohn:
FWIW... that's actually how it works today. You can get telemetry and feedback docs, but you have to set up separate `couch2pg` process. The purpose of this ticket was to make it easier to get that data by allowing users to specify whether or not to include it in one `couch2pg` process so that you _don't_ have to set up a second one. 

RE: future partners only giving access to telemetry data, I'd imagine we would create a specific API (and user privelage) for that rather than using couch2pg.

[...]

PR #63, opened 2021-02-19
Patch 1, Users-meta views

yrimal:
#54


garethbowen:
@yrimal It looks like these views were added in [this commit](https://github.com/medic/medic-couch2pg/commit/9f9d7c3b56b8d68f57e511f719ff1385bb3b966e). Is this PR still needed?

yrimal:
Correct, this is not required @garethbowen .

PR #68, opened 2021-03-02
Replace Travis by GitHub Actions

latin-panda:
# Description

This PR:
- Removes travis file.
- Configure Github Actions
- Update Docker file with supported node version.

**Ticket:** https://github.com/medic/medic-couch2pg/issues/67

# License


Dockerfile
```diff
@@ -10,3 +9,4 @@ RUN npm ci
 RUN npm i -g grunt-cli
 
 COPY . .
+CMD ["node", "index.js"]
```
latin-panda, line 12: Without this it doesn't run in local and doesn't run test.

Dockerfile
```diff
@@ -1,6 +1,5 @@
 ARG node_version
-FROM node:$node_version
-
+FROM node:10
```
latin-panda, line 2: Not working with node 8 but with 10 works fine.

latin-panda:
Hi @craig-landry, can you please take a look at this PR? Thank you!

.github/FUNDING.yml
```diff
@@ -0,0 +1,12 @@
...
+github: medic
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
```
craig-landry, line 12: Do any of these provide value by being here in this empty state? If not, please remove them.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  build-and-test:
+    name: Build medic-couch2pg
+    runs-on: ubuntu-18.04
+
+    steps:
+    - name: Checkout branch
+      uses: actions/checkout@v2
+
+    - name: Use nodeJS 10.x
```
craig-landry, line 18: This is more about setting up, right?
```suggestion
    - name: Setup NodeJS
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: 10.x
```
craig-landry, line 21: The comment at the top notes the "pinned" management of the dependencies. The `.x` here doesn't seem to match that comment. Is the desire for this to dynamically pick up the latest from version 10? If so, that seems fine, but the comment at the top would need to match. If we do want a specific version though, let's just specify that here.... or does this step input not allow for that?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: Set environment variables
+      run: |
+        echo "COUCH2PG_SLEEP_MINS=120" >> $GITHUB_ENV
+        echo "COUCH2PG_DOC_LIMIT=1000" >> $GITHUB_ENV
+        echo "COUCH2PG_RETRY_COUNT=5" >> $GITHUB_ENV
```
craig-landry, line 27: Are these env vars needed in multiple steps? If not, please move them to the step that needs them with the more "normal" `env` syntax. If they are needed in multiple, the duplication is undesirable and I was surprised to see that this is indeed the recommended approach, in which case this is fine as-is.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: Build
+      run: docker-compose build test
+
+    - name: Run unit-test
```
craig-landry, line 32: Minor nitpick... It is preferred that this is a friendly display name:
```suggestion
    - name: Run tests
```

Dockerfile
craig-landry: How did this work in the past?

Dockerfile
craig-landry: Is there a ticket to have versions other than 8 fixed up? If so, please link it here. The lack of 8 support isn't terrible, but failing on 12 is surprising.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
+# Notes:
+#  - All software installed below is pinned so they don't get upgraded and break w/o us knowing why.
+#  - If you bump one version up to latest, please bump them all so we stay current!
+
+name: Build medic-couch2pg and run unit test cases
+
+on: [push, pull_request]
+
+jobs:
```
craig-landry, line 9: The old travis config had a notification step, but I don't see that here. Is that an intentional omission?

craig-landry (changes requested):

Dockerfile
latin-panda: I found this old issue related and with similar error: https://github.com/medic/medic-couch2pg/issues/42
Maybe because of the dependencies in this project. 
I think @njogz might have fixed it for Node 8 (?) by setting a fix version number for 
`"pouchdb-core": "7.0.0",
    "pouchdb-adapter-http": "7.0.0"`

latin-panda: The issue still happening, currently researching...

.github/FUNDING.yml
latin-panda: Done

.github/workflows/ci.yml
latin-panda: Done

.github/workflows/ci.yml
latin-panda: I've fixed the build in node versions in this [PR](https://github.com/medic/medic-couch2pg/pull/73), code adapted to test in version 8, 10 and 12.

Dockerfile
latin-panda: I've fixed the build in node versions in this [PR](https://github.com/medic/medic-couch2pg/pull/73), code adapted to test in version 8, 10 and 12.

Dockerfile
latin-panda: After fixing the build in node 8, 10 and 12 in another PR, the original version of this file is working again.

.github/workflows/ci.yml
latin-panda: Done!

.github/workflows/ci.yml
latin-panda: Fixed! I found a nicer way of setting these inside the step, instead of global

.github/workflows/ci.yml
latin-panda: This is configured in the repo [settings](https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/configuring-notifications#github-actions-notification-options), I'm already getting the emails when it fails :)

latin-panda:
I'm canceling this PR and making a new one because the diff is too confusing. Apologies.

PR #69, opened 2021-03-03
Replicate medic users meta database

njogz:
Adds the ability to replicate medic-users-meta database to PostgreSQL. 

medic/medic-couch2pg#54

### DIFF
--- a/env.js
+++ b/env.js
@@ -17,7 +17,6 @@ module.exports = function() {
     changesLimit: safeNum(process.env.COUCH2PG_CHANGES_LIMIT),
     retryCount: safeNum(process.env.COUCH2PG_RETRY_COUNT),
     sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60),
-    couchdbUsersMeta: process.env.COUCH2PG_USERS_META || false,
     couchdbUsersMetaDocLimit: process.env.COUCH2PG_USERS_META_DOC_LIMIT || 50,
   };
 };
--- a/env.js
+++ b/env.js
@@ -18,5 +18,6 @@ module.exports = function() {
     retryCount: safeNum(process.env.COUCH2PG_RETRY_COUNT),
     sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60),
     couchdbUsersMeta: process.env.COUCH2PG_USERS_META || false,
+    couchdbUsersMetaDocLimit: process.env.COUCH2PG_USERS_META_DOC_LIMIT || 50,
   };
 };
--- a/env.js
+++ b/env.js
@@ -16,6 +16,7 @@ module.exports = function() {
     docLimit: safeNum(process.env.COUCH2PG_DOC_LIMIT),
     changesLimit: safeNum(process.env.COUCH2PG_CHANGES_LIMIT),
     retryCount: safeNum(process.env.COUCH2PG_RETRY_COUNT),
-    sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60)
+    sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60),
+    couchdbUsersMeta: process.env.COUCH2PG_USERS_META || false,
   };
 };
--- a/libs/couch2pg.js
+++ b/libs/couch2pg.js
@@ -1,6 +1,6 @@
 const urlParser = require('url');
 const PouchDB = require('./db');
-const couch2pg = require('couch2pg-njuguna');
+const couch2pg = require('couch2pg');
 
 // Removes credentials from couchdb url
 // Converts http://admin:pass@localhost:5984/couch1
--- a/libs/couch2pg.js
+++ b/libs/couch2pg.js
@@ -10,13 +10,14 @@ const parseSource = url => {
   return `${source.host}${source.path}`;
 };
 
-const replicate = (couchUrl, pgconn, opts) => {
+const replicate = (couchUrl, pgconn, opts, table) => {
   return couch2pg.importer(
           pgconn,
           new PouchDB(couchUrl),
           opts.docLimit,
           opts.changesLimit,
-          parseSource(couchUrl)).importAll();
+          parseSource(couchUrl),
+          table).importAll();
 };
 
 const migrate = (pgUrl) => {
@@ -24,6 +25,6 @@ const migrate = (pgUrl) => {
 };
 
 module.exports = {
-  replicate: (couchUrl, pgconn, opts) => replicate(couchUrl, pgconn, opts),
+  replicate: (couchUrl, pgconn, opts, table = 'couchdb') => replicate(couchUrl, pgconn, opts, table),
   migrate: (pgUrl) => migrate(pgUrl)
 };
--- a/libs/couch2pg.js
+++ b/libs/couch2pg.js
@@ -1,6 +1,6 @@
-const urlParser = require('url'),
-      PouchDB = require('./db'),
-      couch2pg = require('couch2pg');
+const urlParser = require('url');
+const PouchDB = require('./db');
+const couch2pg = require('couch2pg-njuguna');
 
 // Removes credentials from couchdb url
 // Converts http://admin:pass@localhost:5984/couch1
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
   "dependencies": {
     "chalk": "2.4.1",
     "clui": "0.3.6",
-    "couch2pg": "^0.4.1",
+    "couch2pg-njuguna": "^0.5.0",
```
njogz, line 20: Using this package to make it easier to test and not have to wait for the accompanying couch2pg change to me merged and released. It will be changed back to `couch2pg` and updated to the correct version number before merging.

env.js
```diff
@@ -16,6 +16,8 @@ module.exports = function() {
     docLimit: safeNum(process.env.COUCH2PG_DOC_LIMIT),
     changesLimit: safeNum(process.env.COUCH2PG_CHANGES_LIMIT),
     retryCount: safeNum(process.env.COUCH2PG_RETRY_COUNT),
-    sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60)
+    sleepMins: (safeNum(process.env.COUCH2PG_SLEEP_MINS) || 60),
+    couchdbUsersMeta: process.env.COUCH2PG_USERS_META || false,
```
garethbowen, line 20: I think we should remove this environment variable and turn it on for everyone.

env.js
```diff
@@ -16,6 +16,8 @@ module.exports = function() {
...
+    couchdbUsersMetaDocLimit: process.env.COUCH2PG_USERS_META_DOC_LIMIT || 50,
```
garethbowen, line 21: Just use the `docLimit` from above. No need to have a separate variable.

libs/couch2pg.js
```diff
@@ -10,20 +10,21 @@ const parseSource = url => {
...
+          table).importAll();
 };
 
 const migrate = (pgUrl) => {
   return couch2pg.migrator(pgUrl)();
 };
 
 module.exports = {
-  replicate: (couchUrl, pgconn, opts) => replicate(couchUrl, pgconn, opts),
+  replicate: (couchUrl, pgconn, opts, table = 'couchdb') => replicate(couchUrl, pgconn, opts, table),
```
garethbowen, line 28: I think we should remove the default option here and hardcode 'couchdb' into the calling functions so it's explicit what table is being used.

libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
```diff
@@ -0,0 +1,43 @@
...
+SELECT 
+    doc->>'_id' AS uuid,
+    CONCAT(
+    doc#>>'{metadata,year}',
+    '-',
+    CASE WHEN 
+            string_to_array(doc#>>'{metadata,versions,app}','.')::int[] < string_to_array('3.8.0','.')::int[] THEN 
+            (doc#>>'{metadata,month}')::int 
+        ELSE 
+            (doc#>>'{metadata,month}')::int+1 END,
```
garethbowen, line 30: NICE!

libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
```diff
@@ -0,0 +1,43 @@
...
+    doc#>>'{metrics,boot_time,min}' AS boot_time_min,
+    doc#>>'{metrics,boot_time,max}' AS boot_time_max,
+    doc#>>'{metrics,boot_time,count}' AS boot_time_count,
+    doc#>>'{dbInfo,doc_count}' AS doc_count_on_local_db
+FROM
+    couchdb_users_meta
+WHERE
+    doc->>'type'='telemetry';
+
+CREATE UNIQUE INDEX idx_useview_telemetry_period_start_user ON useview_telemetry(period_start,user_name);
```
garethbowen, line 43: Please reach out to Yuvraj to review this schema.

tests/replicate_users_meta.js
```diff
@@ -0,0 +1,40 @@
...
+  it('replicates single couch record to the right table on postgres', async() => {
+    await replicate(couchUrl, pgUrl, {timesToRun: 1, couchdbUsersMeta: true});
+    let rows = await pg.rows('couchdb_users_meta');
+    expect(rows.length).to.equal(1);
+    const [couchRecord, pgRecord] = [singleMedicDoc, rows[0].doc];
+    expect(pgRecord.sex).to.equal(couchRecord.sex);
+    expect(pgRecord.role).to.equal(couchRecord.role);
+    expect(pgRecord.type).to.equal(couchRecord.type);
+  });
+});
```
garethbowen, line 40: Ideally we'd also be testing the migration and materialised view. Do you think it would be easy enough to execute that and query the view to ensure it's outputting the rows we expect?

garethbowen (changes requested):
Nice one! A few comments inline but nothing major.

env.js
njogz: The docs in the users-meta database seem to be quite a bit larger than other docs possibly due to the presence of stack traces. This was causing the app to crash when I tested locally using docLimit so decided to have an additional parameter that can be set lower than the docLimit.

env.js
njogz: The ticket states that it should be an opt-in feature by default rather than it being on for everyone.

tests/replicate_users_meta.js
njogz: Yes. I have added a test for the material views an the migration.

env.js
garethbowen: Ok! Good call. We probably need to solve this for both cases so they automatically recover when this is set too high but that's outside of this issue.

libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
garethbowen: Should this be `userview` not `useview`?

njogz: Good catch. I got the code for the views from Yuvraj so I'll confirm with him.

njogz: `useview` is intentional and not a typo. The pattern the team has been following is `Materialized views: useview_` and `Views: formview_`.

env.js
garethbowen: @michaelkohn and I just discussed this and we've both agreed it should be on for everyone. Sorry for the confusion in the issue!

libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
garethbowen: Ok!

garethbowen (changes requested):
Nice work.

The one last change is to hardcode it to replicate the user meta data rather than having an environment variable.

env.js
njogz: No worries. I have made the change. Please have another look when you can.

garethbowen (approved):
Great work!

The build failed but it looks unrelated to your work so I've restarted it again.

dianabarsan:
I had to rebase to fix conflicts, as a consequence this comment was orphaned, so linking it here so it isn't lost: https://github.com/medic/medic-couch2pg/commit/4e2e044cbe8d9f52fc19ff5ef9321697548cca56#commitcomment-49150811

package.json
dianabarsan: I've published `couch2pg@0.5.0` and reverted this change.

env.js
nomulex: don't we need a safeNum safety check here?

nomulex (commented):
Just curious whether we need to put a check

env.js
garethbowen: @nomulex Good spotting! As this has already been merged can you please raise a new issue for it?

nomulex: Ticket logged [here](https://github.com/medic/cht-couch2pg/issues/99) @garethbowen

PR #73, opened 2021-03-17
Fixing build in node 8, 10 and 12

latin-panda:
**Description**

This PR fix:
- Removes curl_request dependency 
- Replaces curl_request code in unit test to use native node js request and ensure long lasting support through node versions.
- Bump up Jest to v 23 to support Node 12 safely.
- Adds login to Docker in travis file to avoid [pull rate limit](https://www.docker.com/increase-rate-limits) errors.

Ticket: https://github.com/medic/medic-couch2pg/issues/42

**The story**

We're currently using [curl-request](https://www.npmjs.com/package/curl-request) that doesn't support Node.js 12+ and hasn't been maintained in 3 years, looks like a dead project. 
Also, this library has [node_libcurl](https://www.npmjs.com/package/node-libcurl) as dependency that includes support for Node.js 12 in release [2.0.0](https://github.com/JCMais/node-libcurl/releases/tag/v2.0.0) (current release is [2.3.1](https://github.com/JCMais/node-libcurl/releases)) but `curl-request` only supports node_libcurl@1.3.3

`curl-request` is used for testing [to know when the db is ready](https://github.com/medic/medic-couch2pg/blob/master/tests/bootstrap.js#L39).

**License**


package.json
```diff
@@ -30,12 +30,11 @@
   },
   "devDependencies": {
     "chai": "^3.0.0",
-    "curl-request": "1.1.1",
     "grunt": "^1.0.3",
     "grunt-cli": "1.2.0",
     "grunt-mocha-test": "0.13.3",
     "gruntify-eslint": "4.0.0",
-    "jest": "22.4.3",
+    "jest": "23.0.0",
```
latin-panda, line 37: To support Node 12, not breaking tests.

tests/bootstrap.js
```diff
@@ -65,6 +71,22 @@ const waitForPg = async (url) => {
   }});
 };
 
+const isDBReady = (url = '') => {
+  return new Promise((resolve, reject) => {
+    http
```
latin-panda, line 76: Going native, because 1 usage for testing and to ensure long live across node versions.

latin-panda:
Hi @garethbowen, when you have time, can yo please take a look at this possible solution? Thanks!

tests/bootstrap.js
```diff
@@ -1,5 +1,5 @@
-const CurlRequest = require('curl-request'),
-      knex = require('knex');
+const knex = require('knex');
+const http = require('http');
```
garethbowen, line 2: This won't work if the URL is `https` (with the s at the end), and most URLs should be https! There's an `https` module you'll need instead. One of the problems with the native solution is you need to analyse the URL to determine which module to import. So either add that (we have it in some other repos), or use a lib.

garethbowen (changes requested):
Green build, awesome!

One comment inline.

tests/bootstrap.js
latin-panda: Yes, I tried `https` module but since our test is using `http` url then it was failing. I'll add a condition to use the right module in case someone changes the test url to https..maybe with regex 👍

latin-panda:
Thanks for the feedback! @garethbowen I've added the support of https and tested again all 3 node versions, working fine in my local. 
Can you please take a look again? Thanks!

tests/bootstrap.js
```diff
@@ -65,6 +70,24 @@ const waitForPg = async (url) => {
   }});
 };
 
+const isDBReady = (url = '') => {
```
garethbowen, line 73: Don't default unless it actually provides value. In this case you've replaced one error with another, and it'll even keep retrying until the tests timeout (I guess?).

Much better to fail fast here. Add a check that runs in the `before` just before the `waitForCouch` call that throws if either of the required env vars are missing, and then you can assume they're set everywhere without default params.

tests/bootstrap.js
```diff
@@ -36,12 +34,19 @@ const waitForDb = async ({url, fn, retries=0}) => {
 const waitForCouch = async (url) => {
   await waitForDb({url: url, fn: async () => {
     try {
-      const {statusCode} = await curl.get(url);
-      if(statusCode === 200) {
+      const result = await isDBReady(url);
+
+      if (result.ready) {
```
garethbowen, line 39: If I'm reading this correctly, `isDBReady` only resolves when `ready` is true, and otherwise it rejects. This means there's lots of dead code here.

Either change this code to assume `ready` is true when the promise resolves, _or_ change the `isDBReady` to always resolve and never reject.

I think I prefer the second option as it makes the handling code nicer - the server not being up yet is an expected case that we can handle by retrying. We only need the catch for unexpected things!

For example...

```js
const isDBReady = (dbUrl) => {
  const parsed = new url.URL(dbUrl);
  const http = parsed.protocol === 'https:' ? require('https') : require('http');

  return new Promise((resolve, reject) => {
    http
      .get(dbUrl, res => {
        resolve({ ready: res.statusCode === 200 });
      })
      .on('error', (error) => {
        resolve({ ready: false, error });
      });
  });
};
```

tests/bootstrap.js
```diff
@@ -65,6 +70,24 @@ const waitForPg = async (url) => {
   }});
 };
 
+const isDBReady = (url = '') => {
+  const http = url.indexOf('https') > -1 ? require('https') : require('http');
```
garethbowen, line 74: This is a fairly rough way to check for which protocol to use. Consider this instead:

```suggestion
  const urlIMPORTED = require('url'); // put this at the top of the file
  
  const parsed = new urlIMPORTED.URL(url);
  const http = parsed.protocol === 'https:' ? require('https') : require('http');
```

Obviously `urlIMPORTED` is a terrible name so you may way the change the parameter to something else so you can call the package `url`. - I'll leave this up to you!

garethbowen (changes requested):
Awesome work. The build is now much faster and I bet much more stable!

I've put some code readability comments inline.

tests/bootstrap.js
latin-panda: Nice! Done
Screen Shot 2021-03-23 at 2 38 38 pm

tests/bootstrap.js
latin-panda: Ah much better, Done!

tests/bootstrap.js
latin-panda: This is better. Done!

latin-panda:
Hi @garethbowen, I've applied the feedback, can you please take a look? Thank you!

garethbowen (approved):
Awesome, thanks! No AT required for this one - merge when ready.

PR #74, opened 2021-03-19
Adding Github Actions

latin-panda:
**Description**

This PR:
- Removes Travis.
- Configure Github Actions

**Ticket:** https://github.com/medic/medic-couch2pg/issues/67

**License**


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+  build-and-test:
+    name: Build medic-couch2pg
+    runs-on: ubuntu-18.04
+
+    strategy:
+      matrix:
+        node-version: [ 8, 10, 12 ]
+
+    steps:
+      - name: Login to Docker Hub
```
latin-panda, line 15: Hi @garethbowen, I'd  like to add the login to avoid reaching the pull limit of Docker Hub, but it's failing. 
I tried doing the travis way like [this](https://github.com/medic/medic-couch2pg/blob/51b296c61d1737951af46220fb4e732d461bdab3/.travis.yml#L12) but it didn't work. I'm wondering if there's a problem in the secrets @ Repository's settings but I'm not admin to check it. 
Can you please look at the settings? 🙏
The rest is working fine, this is the only piece blocking me.

garethbowen:
@latin-panda I've added username and password from 1pass as GH secrets and re-run the jobs. Back to you!

latin-panda:
@garethbowen Thank you! It's working nice. Can you please review? 🙏
This PR depends on the node versions to build fine (this [PR](https://github.com/medic/medic-couch2pg/pull/73))
Do we AT this one?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: Build
+        run: docker-compose build --build-arg node_version=${{ matrix.node-version }} test
+        env:
+          COUCH2PG_SLEEP_MINS: 120
+          COUCH2PG_DOC_LIMIT: 1000
+          COUCH2PG_RETRY_COUNT: 5
```
garethbowen, line 29: These environment variables look new - are they needed? What changed?

garethbowen (commented):
Looks great so far. One question about environment vars inline...

.github/workflows/ci.yml
latin-panda: These were added in this [commit](https://github.com/medic/medic-couch2pg/commit/471d62f36425aa80385c0c193aa379e68f22e5ea#diff-cbf01e88a71952667baadd3a091707e87cf2e42b0c4cff1d2569af68e134e1fe) (the [README](https://github.com/medic/medic-couch2pg/blob/master/README.md)) and used for supporting some logic around the [runner](https://github.com/medic/medic-couch2pg/blob/1bbab3e3014504eb1699f1e5317d6abf73b8bfb9/libs/medic/runner.js#L55) and the [inquirer](https://github.com/medic/medic-couch2pg/blob/1bbab3e3014504eb1699f1e5317d6abf73b8bfb9/inquirer.js#L63). 
Maybe we can leave it to mirror prod, is that okay?

latin-panda: I can see it runs without these env variables in local.

But asking around, I can confirm it's used in production where:
COUCH2PG_SLEEP_MINS = is the number of minutes between syncs
COUCH2PG_DOC_LIMIT = is the number of documents couch2pg fetches from couchdb everytime it calls couchdb
COUCH2PG_RETRY_COUNT =  is how many times couch2pg should keep trying syncing docs after experiencing an error.

garethbowen: What I meant was, specifying the environment variables in the build is new. I can't see them in the travis file anywhere. Why did you add them here? Are they needed?

latin-panda: Removed the env variables.

latin-panda:
Hi @garethbowen, I've removed the env variables, we can build without them

garethbowen (approved):
Awesome work!

No AT required - merge when ready.

PR #76, opened 2021-03-24
Update 202102191153.do.54-create-couchdb-users-meta-table.sql

yrimal:
Fixed an issue with matview refesh crashing in non SEMVER version.


yrimal:
Pushed to 54 from internal code, so closing this PR as unnecessary.

Issue #87, opened 2021-05-17
`useview_telemetry` is failing for instances running on CHT-Core > 3.8.0


dianabarsan:
Reported error: 
```
[2021-05-17T08:32:50.708Z ERROR]:  { error: date/time field value out of range: "2020-13-1"
    at Connection.parseE (/var/adapter/shared/medic-couch2pg-v3.2.1/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/var/adapter/shared/medic-couch2pg-v3.2.1/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/var/adapter/shared/medic-couch2pg-v3.2.1/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:125:13)
    at Socket.emit (events.js:221:7)
    at addChunk (_stream_readable.js:265:12)
    at readableAddChunk (_stream_readable.js:252:11)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at TCP.onread (net.js:598:20)
  name: 'error',
  length: 294,
  severity: 'ERROR',
  code: '22008',
  detail: undefined,
  hint: 'Perhaps you need a different "datestyle" setting.',
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'SQL statement "REFRESH MATERIALIZED VIEW CONCURRENTLY useview_telemetry"\nPL/pgSQL function refresh_matviews() line 15 at EXECUTE',
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'datetime.c',
  line: '3890',
  routine: 'DateTimeParseError' }
```

It seems like this is related to this migration: 
https://github.com/medic/medic-couch2pg/blob/master/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql#L27

Where, instead of adding +1 when the instance version is lower than 3.8.0, we end up adding +1 when the instance is equal or greater, ending up with 13th month dates.

mrsarm:
Assigning myself because I'll work today in the solution for #86 that affects the same view, and doing separate PRs from each one would imply 2 different migrations scripts, recreating 2 times the same view, because PG does not allow to alter views to just change column definitions.

mrsarm:
Ready for AT, along with the issues #86 and #87, branch `86-87-fix-meta-views`.

[...]

PR #88, opened 2021-05-17
updated useview_telemetry

### LINKED ISSUE #86: Unique index in `useview_telemetry` view by `period_start` and `user_name` cause migration to crash if a user has more than one device
The constraint `idx_useview_telemetry_period_start_user` defined [here](https://github.com/medic/medic-couch2pg/blob/5fd77a8d3eb8c5b237958bce285e4bcafffd96b5/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql#L43) makes each record in the `useview_telemetry` view unique by the columns `period_start` and `user_name`:

```sql
CREATE UNIQUE INDEX idx_useview_telemetry_period_start_user ON useview_telemetry(period_start,user_name);
```

But if a user has more than one devices it will send as many records with the same pair `(period_start,user_name)` as devices each month, this includes users that uses desktop browsers not just one mobile phone, so eg. a supervisor that access the CHT from 2 different computers with the same account will produce 2 records that will be stored in the `medic-users-meta` DB in Couch, then when synchronized with medic-couch2pg the records will collide, raising the following exception that will prevent this app to migrate the data from Couch to Postgres:

```diff
-[2021-05-14T20:33:45.835Z ERROR]:  error: duplicate key value violates unique constraint "idx_useview_telemetry_period_start_user"
    at Connection.parseE (/projects/medic/medic-couch2pg/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/projects/medic/medic-couch2pg/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/projects/medic/medic-couch2pg/node_modules/pg/lib/connection.js:113:22)
    at Socket.emit (events.js:314:20)
    at addChunk (_stream_readable.js:298:12)
    at readableAddChunk (_stream_readable.js:273:9)
    at Socket.Readable.push (_stream_readable.js:214:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
  length: 551,
  severity: 'ERROR',
  code: '23505',
+  detail: 'Key (period_start, user_name)=(2021-05-01, adama) already exists.',
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'SQL statement "INSERT INTO public.useview_telemetry SELECT (diff.newdata).* FROM pg_temp_2.pg_temp_16536_2 diff WHERE tid IS NULL"\n' +
    'SQL statement "REFRESH MATERIALIZED VIEW CONCURRENTLY useview_telemetry"\n' +
    'PL/pgSQL function refresh_matviews() line 15 at EXECUTE statement',
  schema: 'public',
  table: 'useview_telemetry',
  column: undefined,
  dataType: undefined,
  constraint: 'idx_useview_telemetry_period_start_user',
  file: 'nbtinsert.c',
  line: '414',
  routine: '_bt_check_unique'
}
[2021-05-14T20:33:45.836Z ERROR]:  An unrecoverable error occurred
[2021-05-14T20:33:45.836Z ERROR]:  TypeError: Assignment to constant variable.
    at Object.run (/projects/medic/medic-couch2pg/libs/medic/runner.js:55:20)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[2021-05-14T20:33:45.836Z ERROR]:  exiting
```

Even if it's unlikely this to happen, only bast with 1 user accessing from more than one device to make the whole process to fail.

I think the best solution is to patch the app removing the index, because even adding to the index something like the device id, the error is still likely to happen soon or later, because there are cases where a device can yield more than 1 aggregation with the same device within the same period.

If it's useful to query this data by user, we can change the index removing the unique constraint.

The solution should also take into account https://github.com/medic/medic-couch2pg/issues/85.

kitsao:
Resolves #86

### DIFF
--- a/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
+++ b/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
@@ -17,6 +17,8 @@ WHERE
 
 CREATE UNIQUE INDEX idx_useview_feedback_period_start_user ON useview_feedback(period_start,user_name);
 
+DROP INDEX IF EXISTS idx_useview_telemetry_period_start_user;
+DROP MATERIALIZED VIEW IF EXISTS useview_telemetry;
 CREATE MATERIALIZED VIEW useview_telemetry AS
 SELECT 
     doc->>'_id' AS uuid,
@@ -40,4 +42,5 @@ FROM
 WHERE
     doc->>'type'='telemetry';
 
-CREATE UNIQUE INDEX idx_useview_telemetry_period_start_user ON useview_telemetry(period_start,user_name);
+CREATE UNIQUE INDEX idx_useview_telemetry_uuid ON useview_telemetry(uuid);
+CREATE INDEX idx_useview_telemetry_period_start_user ON useview_telemetry(period_start, user_name);
--- a/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
+++ b/libs/medic-users-meta/migrations/202102191153.do.54-create-couchdb-users-meta-table.sql
@@ -25,12 +25,10 @@ SELECT
     CONCAT(
     doc#>>'{metadata,year}',
     '-',
-    CASE WHEN 
-            string_to_array(substring(doc#>>'{metadata,versions,app}' FROM '(\d+.\d+.\d+)'),'.')::int[] < '{3,8,0}'::int[] THEN 
-            (doc#>>'{metadata,month}')::int 
-        ELSE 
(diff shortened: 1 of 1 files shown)


mrsarm:
Hey @derickl , I going to create another PR to supersede this one, I discussed the approach with @kitsao . In the new PR will address also #87 . Moreover, the patch need to be applied in a new migration file, not in existent ones that are inmutable (check https://github.com/medic/medic-couch2pg/issues/78#issuecomment-827952017).

I already have the fix but I found one more error that I going to address also in the same PR (I'll create a new issue with the description before the PR).

mrsarm:
Superseded by https://github.com/medic/medic-couch2pg/pull/91.

Issue #82, opened 2021-05-07
Add support for later node versionz


ngaruko:
At the moment, apparently `medic-couch2pg` does not work with node versions greater that 10.
With the v14.16.0 , I couldn't get the node . command to do anything. No error in the logs just the first json.
![image](https://user-images.githubusercontent.com/6979995/117397056-63687780-af4f-11eb-87c1-54db74deef5d.png)

mrsarm:
2 more things that are very related to this:

1. Dependencies are very outdated, with 45 vulnerabilities, 2 of them critical:

   ```diff
   $ npm ci
   
   added 953 packages, and audited 954 packages in 25s
   
   -45 vulnerabilities (26 low, 2 moderate, 15 high, 2 critical)
   ```

   The library `node-libcurl` that cause the error in some versions of node / platforms is one the outdated libraries so upgrading it for sure will solve the problem.

2. We released a new version of couch2pg around a month ago: [v.0.6.0](https://github.com/medic/couch2pg/releases/tag/v0.6.0), this project relies on it but we are using instead the v0.5.0.

@garethbowen can we prioritize this so next release not only is a rebranding from ~medic~cht-couch2pg but we also fix this that my guess is that will only require to execute `npm audit` and upgrade the version of couch2pg?

garethbowen:
I think couch2pg 0.5 and 0.6 are identical due to a glitch during publishing, so I don't think bumping that version will do anything. In general we should aim to bump dependencies every release at least for non-breaking changes, so it'd be good to do this, potentially for both libraries at the same time.

PR #96, opened 2021-05-25
Switch to cht-couch2pg

dianabarsan:
medic/cht-core#6586

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# medic-couch2pg
+# CHT-couch2pg
 
 Software for creating read-only replicas of CouchDB data inside PostgreSQL v9.4.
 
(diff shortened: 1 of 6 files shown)


dianabarsan:
hey @mrsarm 
I think I got everything ... I left the `bin` entry unchanged because the commit that added it intended additional entries to be added per database, and our default database is still called "medic". 
Please review. Thanks!

README.md
```diff
@@ -1,4 +1,4 @@
-# medic-couch2pg
+# CHT-couch2pg
```
mrsarm, line 1: It doesn't feel good "CHT" in uppercase here followed by `-` and then the other part in lowercase and without spaces for a title, instead of `cht-couch2pg` or `CHT - couch2pg`, but is not wrong, just an opinion. Specially for a CLI tool, I would prefer `cht-couch2pg` (all lower case), and checking one of the examples @garethbowen mention in the ticket:

> ... Medic Gateway -> CHT Gateway

My be worth to follow a pattern for all the renaming PRs, so, beside the decision of renaming ~medic-~ → _cht-_ in the CLI names and project URLs, do we discussed how to rename the product names?

package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "medic-couch2pg",
-  "version": "3.3.0",
-  "description": "replicate medic data to postgres database",
+  "name": "cht-couch2pg",
+  "version": "3.4.0",
```
mrsarm, line 3: Are we going to do an immediate release? I would like to address first #82 , I'm trying to prioritize it but maybe it's better in the meantime to move the version change in a new PR once we decide to release.

dianabarsan: I think we can release immediately. 
Since it will be a different npm package, we're not worried someone might upgrade accidentally.

README.md
dianabarsan: I interpreted the section as a "project name", so that's why I used caps. 
image

I agree the cli names should be lowercase, and it's how I changed the other instances:
image

garethbowen: Maybe just remove the hyphen from the title?

package.json
```diff
@@ -1,12 +1,12 @@
...
+  "description": "replicate cht data to postgres database",
```
garethbowen, line 4:
```suggestion
  "description": "Replicate CHT data from CouchDB to postgres database",
```

garethbowen (approved):
Nice. Two minor suggestions inline.

PR #100, opened 2021-07-06
add documentation on COUCH2PG_USERS_META_DOC_LIMIT and add safeNum safety

njogz:
Add documentation about the `COUCH2PG_USERS_META_DOC_LIMIT` environment variable and add safeNum safety check

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ The supported environment variables are:
 | -------------------- | ---------------------------------------------------------------------------------------------------------- |
 | POSTGRESQL_URL       | PostgreSQL instance URL, format: `postgres://[user]:[password]@localhost:[port]/[database name]`           |
 | COUCHDB_URL          | CouchDB instance URL, format: `https://[user]:[password]@localhost:[port]/medic`                           |
-| COUCH2PG_SLEEP_MINS  | Number of minutes between synchronization                                                                  |
+| COUCH2PG_SLEEP_MINS  | Number of minutes between synchronization. It defaults to 50 if this variable is not explicitly set.       |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,6 +23,7 @@ The supported environment variables are:
 | COUCH2PG_SLEEP_MINS  | Number of minutes between synchronization                                                                  |
 | COUCH2PG_DOC_LIMIT   | Number of documents cht-couch2pg fetches from CouchDB everytime                                          |
 | COUCH2PG_RETRY_COUNT | Number of times cht-couch2pg will retry synchronizing documents from CouchDB after experiencing an error |
+| COUCH2PG_USERS_META_DOC_LIMIT | Number of documents from users-meta database that cht-couch2pg fetches from CouchDB everytime |
```
mrsarm, line 26: The description seems incomplete, making the variable to look like it sets a limit on how much documents are fetched in total or in total in a batch, while in reality the value is set to limit the number of documents *per* fetch. A good description I found is [here](https://github.com/medic/cht-couch2pg/blob/3e38727075c0132fad3f3e41c4a4461b202519de/inquirer.js#L65), and would be nice to also add in the description what is the default value taken when the variable is not set.

Also would be nice if you can add in the documentation what is the default value taken in the rest of the variables or at least where to check the default values _(inquirer.js)_.

mrsarm (changes requested):
Hey, thanks for taken this! I also linked the PR to another issue that requested the same changes.

mrsarm (approved):
LGTM :+1:

yrimal (approved):

Issue #56, opened 2020-02-20
Architecture v3: couch2pg + postgres


Hareet:
To satisfy: https://github.com/medic/cht-core/issues/6189

We should initialize Postgres with custom parameters for 1 project.

- CI publishes couch2pg docker image
- docker-compose consists of adding couch2pg + a postgres container to existing medic-net network
- initialization setups up postgres accurately

Hareet:
@yrimal 
There's a lot of people who access our Postgres Server. I want to make sure all the use-cases still function or can be adequately substituted while this is re-architected in a different network. 

Can you describe what you are doing when you access the server (crons, scripts) related to postgres, and what's typical use for partners, since I see a lot of partner type logins present? 

Thanks

yrimal:
Hi @Hareet ,
For impact monitoring, impact team creates certain views and functions on each database. It's not always the same query for all database so we can't containerize it. Once impact functions have been installed on these databases, we've a dblink connection set from `dwh_impact` database that queries every databases to pull updated data every other night. 

Apart from this, when there's a need to query database for other purposes, then also we need to access those databases.

yrimal:
For partner users, only some (probably just lg) have direct access to the database. But, at least one account will still need access from klipfolio for partner facing dashboards. If things are not clear, I can talk over call as well.

Hareet:
Testing:
- [ ] Be able docker-compose medic-os container or arch v3 containers (docker-compose)
- [ ] Launch couch2pg + postgres containers after medic-os is initiliazed, and verify data is inserted to postgres
- [ ] Entrypoints work in restarts, crashes and initial setup

Hareet:
We have labeled this ticket as 5 points for SRE Sprint 2

Unable to share medic-infrastructure labels with this repo.

Hareet:
completed!

PR #101, opened 2021-07-09
Arch v3

nomulex:
Addresses  Issue #56

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,4 @@ RUN npm i -g grunt-cli
 
 COPY . .
 
-CMD ["/bin/sh", "/usr/app/couch2pg-entrypoint.sh main"]
+CMD ["/bin/sh", "/usr/app/couch2pg-entrypoint.sh", "main"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
-ARG node_version
+ARG node_version=10
 FROM node:$node_version
+
+RUN apt-get update
+RUN apt-get -y install postgresql-client curl
 
 WORKDIR /usr/app
 
@@ -10,3 +13,5 @@ RUN npm ci
 RUN npm i -g grunt-cli
 
 COPY . .
+
+CMD ["/bin/sh", "/usr/app/couch2pg-entrypoint.sh"]
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ NB: The integration tests destroy and re-create the given databases each time th
 
 We support PostgreSQL 9.4 and greater. The user passed in the postgres url needs to have full creation rights on the given database.
 
+## Tesing the entrypoint script together with the npm tests
+
+You can test both the code in the repository and the docker image entrypoint by runing the `docker-compose.test.yml` file. You first need to install the test submodules that are found in in the bash tests folder by running `git submodule update --init`.  Once done you can run the tests by running `docker-compose -f docker-compose.test.yml up -d` and checking the tests results through `docker logs cht-couch2pg_sut_1`
+
 
 ## Example usage
 
@@ -97,12 +101,12 @@ script
 end script
 ```
  - The service is then a standard service, e.g. `service couch2pg-example-client start`
- 
+
 ### Installing as a service using Systemd (18.04.3 LTS [Bionic Beaver])
 
 To setup couch2pg using systemd is also pretty simple. You will need to have sudo rights to the server and then follow the steps listed below:
- 
- - Install git and clone this repo onto your server, check out the relevant tag `git checkout tag_id`, and run `npm ci`. 
+
+ - Install git and clone this repo onto your server, check out the relevant tag `git checkout tag_id`, and run `npm ci`.
  - Create a systemd unit file for your project `sudo` create `/etc/systemd/system/couch2pg-sample-client.service`
  - As we are going to put passwords in this file, you want to `sudo chmod o-r /etc/systemd/system/couch2pg-sample-client.service` so that only root can read it.
  - Edit this file and configure the couch2pg system unit. It could be something simillar to this;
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -74,11 +74,62 @@ NB: The integration tests destroy and re-create the given databases each time th
...
+           POSTGRES_USER_NAME: cht_couch2pg
+           POSTGRES_DB_NAME: cht
+           POSTGRES_PASSWORD: couch2pg_password
+           API_URL: http://medic-api:5988
+        depends_on:
+          - cht-postgres
+  cht-postgres:
+        container_name: cht-postgres
+        image: medicmobile/cht-postgres:release-postgres13-rc.1
+        environment:
```
newtewt, line 85: When I do testing using this. I expose the ports for postgres so that I can connect with PGAdmin.

docker-compose.yml
```diff
@@ -1,26 +1,90 @@
...
+      networks:
+        - cht-net
+      depends_on:
+        - couchdb
+      restart: unless-stopped
+    couch2pg:
+        container_name: cht_couch2pg
+        image: medicmobile/cht-couch2pg:test-rc.1
+        environment:
+           COUCHDB_URL: "http://${COUCHDB_USER:-cht}:${COUCHDB_PASSWORD:-cht_password}@medic-api:5988/medic"
```
newtewt: Can we go directly against the couchdb container instead of through API?

newtewt (commented):
I was making some review notes and it seems like you are pushing changes at the same time. I'll hold off for now.

nomulex:
I have made changes to the way we build and test images.

nomulex:
^^ Upload to docker hub fails because we need to have proper credentials to docker hub. Working with @garethbowen to get this fixed.

couch2pg-entrypoint.sh
```diff
@@ -0,0 +1,66 @@
...
+      echo "No  PostgreSQL DB Found" >&2
+      exit 1
+    fi
+    sleep $SLEEP_SECONDS
+  done
+echo "Postgres is ready moving on ...">&2
+}
+
+
+check_if_api_is_ready(){
```
Hareet, line 33: Related to one of Nick's comments, should this be: `check if couchdb is ready` or a mix of both? The flow is, couchdb is required here, but if you set up the endpoint as API, then API would check if couchdb is available.

docker-compose.yml
```diff
@@ -1,26 +1,41 @@
...
-      COUCHDB_USER: ${COUCH_ADMIN}
-      COUCHDB_PASSWORD: ${COUCH_PASS}
+      COUCHDB_USER: cht
+      COUCHDB_PASSWORD: cht-password
+    networks:
+      - cht-net
   postgres:
-    image: postgres:9.4
-    ports:
-      - ${PG_PORT}:5432
```
Hareet, line 13: let's expose this port again to resolve Nick's comment above. If there's a reason not to, let's reply and add some documentation to the README on how to expose if required.

README.md
```diff
@@ -33,122 +33,137 @@ export COUCH2PG_DOC_LIMIT=1000
...
 
+#### 9. POSTGRES_DB_NAME
+
+The name of the PostgreSQL database to sync to.
+
+#### 10. POSTGRES_PASSWORD
+
+The password the cht-couch2pg uses to authenticate to PostgreSQL.
+
+A sample docker-compose snippet is shown below. This image can work with the [cht-postgres](https://hub.docker.com/r/medicmobile/cht-postgres) docker image also available on docker hub.
```
Hareet, line 107: This snippet below, and the cht-postgres image isn't the same one used in the docker-compose.yml file. If its possible, I think we should use the upstream postgres image, and mount our configuration or pass our necessary parameters and user creations during the cli command or custom entrypoint. This would also address Aidan's feedback in Slack regarding using upstream images where possible.

Hareet (changes requested):
Thanks so much for this work @nomulex !!
I left a few comments and after those are resolved, this looks good to go!

While you are in the branch, resolving those suggestions, here's a few pieces that will help us organize the configuration files:
- Let's rename the docker-compose.yml to docker-compose-test.yml and add the necessary `-f docker-compose.test.yml` parameter to all docker cli commands / instructions
- Let's add a file named `docker-compose.prod.yml` that has hardcoded the latest production release, and appropriate documentation.

Follow-ups after this PR is merged:
- Create a GH ticket and merge k8s templates for these images

README.md
```diff
@@ -33,122 +33,137 @@ export COUCH2PG_DOC_LIMIT=1000
...
-```
-docker-compose build --build-arg node_version=[node version] test
-docker-compose up
+```bash
+docker-compose build cht-couch2pg
+docker-compose run cht-couch2pg grunt test
 ```
 
-Then in another terminal:
+Run tests in interactive watch mode with: `docker-compose run test npm run watch`
```
Hareet, line 53: This command needs to be updated to call the correct service name and command. Currently returns:
```
docker-compose run test npm run watch
ERROR: No such service: test
```

Hareet (commented):
Requesting another change regarding how to launch tests with verbose/interactive output as well as launch single tests.

Dockerfile
```diff
@@ -1,12 +1,23 @@
-ARG node_version
-FROM node:$node_version
-
-WORKDIR /usr/app
-
-COPY package.json .
-COPY package-lock.json .
-
+#base Build
+ARG node_version=10
```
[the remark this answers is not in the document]
nomulex: Actually, that's a [default value](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact) in case the actual is not provided.   It is more like a fail-safe mechanism.   Docker build uses the ARG if provided and if not it uses the default.

Hareet (approved):
Awesome sauce! Looks great!!

Issue #104, opened 2021-08-06
Time view refreshing and log


aidan-plenert-macdonald:
Time how long it takes to refresh views and time it for benchmarking performance

https://github.com/medic/cht-couch2pg/blob/master/libs/analytics/migrations/201606200952.do.2318-prepareContacts.sql#L93

PR #105, opened 2021-08-06
Benchmark view refreshes #104

aidan-plenert-macdonald:
https://github.com/medic/cht-couch2pg/issues/104

To add statistics to logs so we can see how long queries take.


mrsarm:
@aidan-plenert-macdonald because the way migrations work , all the scripts under the `migrations/` folders are immutable. When we start a new deployment, each migration script is executed one after another, and if it is an upgrade, cht-couch2pg takes care of only run the missed migrations, so any new change need to be implemented in a new script file.

This may look like inefficient because you have to delete the object and re-define it from scratch even if only one line of the whole definition is changed, but it is actually how Postgres works, there is no way to alter just a small part of a function definition , you have to destroy it and then create it again with the changes.

PR #114, opened 2022-02-28
denote required node version in readme

mrjones-plip:
per #medic/couch2pg/issues/31


kitsao (approved):
Looks good.

mrjones-plip:
Thanks @kitsao ! 

@garethbowen  - it looks like @kitsao  doesn't have write access to this repo so the "lgtm" doesn't count I guess? Doesn't seem correct, but maybe there's some history about perms to this repos I'm not aware of.

garethbowen (approved):
LGTM2

Issue #112, opened 2022-02-15
Make sure sample docker_compose works with a remote installation of CHT-Core


dianabarsan:
This came in through the forum: https://forum.communityhealthtoolkit.org/t/read-only-replica-of-couchdb-docs-into-postgres-for-data-visualization/1710/6

I found a few problems with our sample from the readme and some of the docker setup: 

1. It adds a CouchDb container (presumably to use as a source) when cht-couch2pg is supposed to get data out of a CHT instance. Incidentally, the container throws a lot of errors, obfuscating what is actually going on. 
My proposed solution is to remove the CouchDb container entirely and have the sample connect to a CHT instance CouchDb

2. If I set my `COUCHDB_URL` env to be a remote CHT instance (for example: `https://admin:pass@192-168-1-253.my.local-ip.co/medic`), the entry script ends up in an infinite retry loop because it expects a 200 response code. API responds with a 302 on the non-slash ending URL. This can be worked around with setting the URL to `https://admin:pass@192-168-1-253.my.local-ip.co/medic/`, but this seems like kind of a hack. 
My proposed solution is to not [grep specifically for 200](https://github.com/medic/cht-couch2pg/blob/main/couch2pg-entrypoint.sh#L36), but for a generic success status or change the curl command to follow redirects. 

3. After doing the above, I ran into SSL issues with the `my.local-ip.co` cert. 
Specifically: 
```
cht-couch2pg  | https://admin:pass@192-168-1-253.my.local-ip.co/medic
cht-couch2pg  | Set postgres URL to postgres://cht_couch2pg:couch2pg_password@cht-postgres:5432/cht
cht-couch2pg  |   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
cht-couch2pg  |                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
cht-couch2pg  | curl: (60) SSL certificate problem: certificate has expired
cht-couch2pg  | More details here: https://curl.haxx.se/docs/sslcerts.html
cht-couch2pg  | 
cht-couch2pg  | curl performs SSL certificate verification by default, using a "bundle"
cht-couch2pg  |  of Certificate Authority (CA) public keys (CA certs). If the default
cht-couch2pg  |  bundle file isn't adequate, you can specify an alternate file
cht-couch2pg  |  using the --cacert option.
[...]

PR #118, opened 2022-03-10
update docs to have same structure, call out which directory to be in

mrjones-plip:
per #112


mrjones-plip:
@jkuester  - docs only changes to be more explicit about where you should be.  Add parity to enumeration in instructions. Learn how to spell "container" ;)

jkuester (approved):

Issue #119, opened 2022-03-10
Tags and Commits to `main` not pushing new images to Docker Hub


mrjones-plip:
The reason commits to `main` are failing is because we're checking for the `master` branch [in CI](https://github.com/medic/cht-couch2pg/blob/9117634f5720b1401dea9b2c7bdacb54ea114ce0/.github/workflows/ci.yml#L56) instead of `main`.  When we fix this, we should also fix the [name of the image in the compose file](https://github.com/medic/cht-couch2pg/blob/9117634f5720b1401dea9b2c7bdacb54ea114ce0/docker-compose.yml#L21) to be like `main-node-10` as the new images won't use the old repo default brand like they used to with `master-node-10`

Also, tags that are added are not pushing a new image either - this is a mystery as to why?! Here's an [example of this failing](https://github.com/medic/cht-couch2pg/actions/runs/1965865170).

mrjones-plip:
@eljhkrr - Feel like doing one more round of user acceptance testing?  Now that we know the docker image works, this PR publishes it to docker hub and updates the compose file to use the now renamed image.

To test:
1. `git checkout main&&git pull origin`  to get latest code.  You can check you got the latest compose file with: `grep 'medicmobile/cht-couch2pg:main-node-10' docker-compose.yml` (should show 1 line)
2. Run the same command as before with, using your `COUCHDB_URL` value instead of mine: `COUCHDB_URL=https://medic:password@192-168-68-26.my.local-ip.co:8442/medic docker-compose up`

thanks thanks!

mrjones-plip:
@Hareet - Do you know how tags are supposed to work? I tried [pushing a tag and it failed](https://github.com/medic/cht-couch2pg/actions/runs/1971640319) :( 

if you're not worried about it, no biggie - i'll just close this ticket b/c pushes to `main` [now generate new images](https://hub.docker.com/layers/medicmobile/cht-couch2pg/main-node-10/images/sha256-307c1837778a72fc482906626c05e6bb4b6f0014de8172f204f49bf09d3667eb?context=explore) per [this PR](https://github.com/medic/cht-couch2pg/pull/121).

eljhkrr:
Confirmed it's working well.

mrjones-plip:
Awesome - thanks @eljhkrr !

mrjones-plip:
I dug up the Docker Hub admin credentials in the password safe and manually pushed the readme change there.  

[...]

PR #121, opened 2022-03-11
switch CI and docker compose to use main vs master

mrjones-plip:
This PR one problem and then accounts for the fix:
* Change CI to not look for `master` but look for `main` instead.  This means that the the tag pushed to docker hub will be `main-node-VERSION` instead of `master-node-VERSION` (eg `main-node-10` vs `master-node-10`)
* Change docker-compose file to use new nomenclature for hub image name

 per #119

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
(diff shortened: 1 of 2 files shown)


mrjones-plip:
@jkuester  - when you get a sec - third PR will be the charm I think :pray:

jkuester (approved):
:rocket:

Issue #115, opened 2022-03-02
useview_telemetry can't be extended to support all metrics


kennsippell:
The [current useview_telemetry table](https://github.com/medic/cht-couch2pg/blob/6264ffc59cc5fe0c00a6bab1aed0ad348676da43/libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql#L48) supports about 7 metrics in a wide-table format.  If you try to extend this to all metrics, you'll see that many metrics have dynamic names:

For example:

* `replication:<database>:<direction>:docs`
* `user_settings:language:<language_code>`
* `enketo:reports:<form>:<action>:<component>`

Especially with values like "form" in the metric name, it is impossible to extend this view to include all metrics.
We should replace this view with a "tall table"

kennsippell:
This was a quick query I threw together to do an an analysis of form/enketo telemetry metrics

```
with telemetry_metrics as (
  select
    distinct metric
  from (
    select 
      jsonb_object_keys(doc -> 'metrics') as metric
    from couchdb_users_meta
    WHERE (doc ->> 'type') = 'telemetry'
  ) T
),

select
  doc #>> '{metadata,user}'::text[] AS user_name,
  concat_ws(
    '-'::text, doc #>> '{metadata,year}'::text[],
    CASE
        WHEN string_to_array("substring"(doc #>> '{metadata,versions,app}'::text[], '(\d+.\d+.\d+)'::text), '.'::text)::integer[] < '{3,8,0}'::integer[] THEN ((doc #>> '{metadata,month}'::text[])::integer) + 1
        ELSE (doc #>> '{metadata,month}'::text[])::integer
    END,

    CASE
        WHEN (doc #>> '{metadata,day}'::text[]) IS NOT NULL THEN doc #>> '{metadata,day}'::text[]
        ELSE '1'::text
    END
  )::date AS period_start,
  telemetry_metrics.metric,
  jsonb_extract_path(doc -> 'metrics', telemetry_metrics.metric, 'min'::text) as min,
  jsonb_extract_path(doc -> 'metrics', telemetry_metrics.metric, 'max'::text) as max,
  jsonb_extract_path(doc -> 'metrics', telemetry_metrics.metric, 'sum'::text) as sum,
  jsonb_extract_path(doc -> 'metrics', telemetry_metrics.metric, 'count'::text) as count,
[...]

PR #130, opened 2023-04-14
Improved useviews for telemetry metrics and telemetry devices

kennsippell:
#115

`useview_telemetry_metrics` uses a long-format with one row per telemetry metric with columns for min, max, sum, count, sumsqr. Provides index columns for device_id, telemetry_doc_id, period_start, and user_name.

`useview_telemetry_devices` displays the latest device information for each { deviceId, username } pair. Allows for easily lookup of latest state for a user-device.

Related: https://forum.communityhealthtoolkit.org/t/making-changes-to-cht-couch2pg-view-useview-telemetry/2490/3

### DIFF
--- a/libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
+++ b/libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
@@ -5,6 +5,8 @@ TABLESPACE pg_default AS
 
 SELECT
   DISTINCT ON (doc #>> '{metadata,deviceId}', doc #>> '{metadata,user}')
+  
+  doc #>> '{_id}' AS telemetry_doc_id,
   doc #>> '{metadata,deviceId}' AS device_id,
   doc #>> '{metadata,user}' AS user_name,
   
@@ -29,7 +31,6 @@ SELECT
   )::date AS period_start,
     
 
-  doc #>> '{_id}' AS telemetry_doc_id,
   doc #>> '{device,deviceInfo,hardware,manufacturer}' AS device_manufacturer,
   doc #>> '{device,deviceInfo,hardware,model}' AS device_model,
   
--- a/libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
+++ b/libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
@@ -1,0 +1,47 @@
+DROP MATERIALIZED VIEW IF EXISTS useview_telemetry_devices;
+
+CREATE MATERIALIZED VIEW public.useview_telemetry_metrics
+TABLESPACE pg_default AS 
+
+WITH telemetry_docs_with_metric_blob AS (
+ SELECT
+    doc #>> '{metadata,deviceId}' AS device_id,
+    doc #>> '{_id}' AS telemetry_doc_id,
+    doc #>> '{metadata,user}' AS user_name,
+    concat_ws(
+      '-', doc #>> '{metadata,year}',
+      CASE
+          WHEN (doc #>> '{metadata,day}') IS NULL AND ((doc #>> '{metadata,versions,app}') IS NULL OR string_to_array("substring"(doc #>> '{metadata,versions,app}', '(\d+.\d+.\d+)'::text), '.'::text)::integer[] < '{3,8,0}'::integer[]) THEN ((doc #>> '{metadata,month}')::integer) + 1
+          ELSE (doc #>> '{metadata,month}')::integer
+      END,
+      CASE
+          WHEN (doc #>> '{metadata,day}') IS NOT NULL THEN doc #>> '{metadata,day}'
+          ELSE '1'::text
+      END
+    )::date AS period_start,
+    jsonb_object_keys(doc -> 'metrics'::text) AS metric,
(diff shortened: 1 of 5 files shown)


kennsippell:
Hey @michaelkohn I see you're doing some work with telemetry [here](https://forum.communityhealthtoolkit.org/t/telemetry-questions-about-the-contacts-tab-search/2651/3). Would you be interested to test these reusable views for telemetry? I find them easier to work with but would value your feedback.

michaelkohn:
Oh ya, i just ran one of these and it's awesome, sooooper helpful. Will add some review comments.

libs/medic-users-meta/migrations/202304140054.do.telemetry_metrics.sql
```diff
@@ -0,0 +1,47 @@
...
+
+WITH telemetry_docs_with_metric_blob AS (
+ SELECT
+    doc #>> '{metadata,deviceId}' AS device_id,
+    doc #>> '{_id}' AS telemetry_doc_id,
+    doc #>> '{metadata,user}' AS user_name,
+    concat_ws(
+      '-', doc #>> '{metadata,year}',
+      CASE
+          WHEN (doc #>> '{metadata,day}') IS NULL AND ((doc #>> '{metadata,versions,app}') IS NULL OR string_to_array("substring"(doc #>> '{metadata,versions,app}', '(\d+.\d+.\d+)'::text), '.'::text)::integer[] < '{3,8,0}'::integer[]) THEN ((doc #>> '{metadata,month}')::integer) + 1
```
michaelkohn, line 14: For some reason I thought I've seen poorly named branches that don't have version they were based off of in the name. If that's not possible, ignore this comment. If that sounds possible, read-on.

If `'{metadata,versions,app}'` is a poorly named branch, I think the `ELSE` will catch it and just use `'{metadata,month}'` without incrementing, which will be correct sometimes and incorrect sometimes, depending on if that branch was before or after 3.8.0. At this point we really have no clue though. One possible solution would be to just make sure we never have a 0 month, even though it might not be right. The other option is to let the calling query fail so that the person trying to use this matview knows there are some inconsistencies and deals with it in their calling query. Bad branch names are maybe rare enough (or non-existent) that it's not worth dealing with? In those scenarios, you could have non-zero months that are not right but won't fail anyway. ¯\_(ツ)_/¯

michaelkohn (commented):
I didn't run the matview creation scripts anywhere but I did test the underlying query (for the metrics one). I'd like to check out the `_devices` query too.

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
```diff
@@ -0,0 +1,47 @@
+DROP MATERIALIZED VIEW IF EXISTS useview_telemetry_devices;
+
+CREATE MATERIALIZED VIEW public.useview_telemetry_metrics
```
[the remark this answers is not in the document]
kennsippell: oops! :)

libs/medic-users-meta/migrations/202304140054.do.telemetry_metrics.sql
kennsippell: this is just so messy. i'm not sure what to do.

michaelkohn: Actually, I just realized the `period_start` field is being cast to a date in the matview anyway so it'll just blow up here if it comes across a month = 0, not later down the road.

Anyway, yeah. `purge-specfic-tasks-3.9.x` is a branch in one project, though I think in that case it should actually be fine because it should have the correct month index since it is > 3.8. 

Either way, it's really rare, anything pre 3.8 is going to be really old data and probably not useful for analysis anyway, so personally I think it's probably just best to make sure it doesn't crash.

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
```diff
@@ -0,0 +1,53 @@
+DROP MATERIALIZED VIEW IF EXISTS useview_telemetry_devices;
+
+CREATE MATERIALIZED VIEW public.useview_telemetry_devices
+TABLESPACE pg_default AS 
+
+SELECT
```
michaelkohn, line 6: I know we don't have `metrics` in the devices stuff but I'd probably still try to make the columns align as closely as possible with the other query, at a minimum, putting `telemetry_doc_id` first.

michaelkohn (approved):
Added 2c but don't feel strong about spending it so... feel free to merge!

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
kennsippell: [here](https://github.com/medic/cht-couch2pg/pull/130#discussion_r1166860748) you say you like to put the unique identifier first. In this case, the unique identifier is `{ device_id, username }`. Do you have a preference as-to which of these is more important?

michaelkohn: Yeah, but since the unique identifier of the couch doc is the `telemetry_doc_id`, that's the field I like to put first. So to me, it satisfies both of my preferences (unique first AND alignment).

Issue #132, opened 2023-04-27
GitHub actions fail due to "does not have a Release file"


kennsippell:
```
Reading package lists...
E: The repository 'http://security.debian.org/debian-security stretch/updates Release' does not have a Release file.
E: The repository 'http://deb.debian.org/debian stretch Release' does not have a Release file.
E: The repository 'http://deb.debian.org/debian stretch-updates Release' does not have a Release file.
The command '/bin/sh -c apt update' returned a non-zero code: 100
Service 'cht-couch2pg' failed to build : Build failed
```

kennsippell:
Reproduces when running `docker run -it node:10 apt update`

PR #133, opened 2023-04-27
Use node:10-buster images to workaround "apt update" failures

kennsippell:
This PR changes the docker image in the `Dockerfile` from `node:10` to `node:10-buster` 

Fix for #132

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on: [ push]
 jobs:
   build-and-test:
     name: Build cht-couch2pg
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 15
 
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on: [ push]
 jobs:
   build-and-test:
     name: Build cht-couch2pg
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mrjones-plip (approved):
Nice job on a quick fix!

I think this PR is fine as is, approving and then leaving a few comments for you to decide if they're actionable.  Happy to re-review if there's interest!

* Today's Node 10 image has  [8 Critical vulnerabilities](https://hub.docker.com/layers/library/node/10.24.1/images/sha256-646d94ee1cdf31632c4f4fc5b2d7567f5484fb4d8bca21ddbea161775354cd59?context=explore)  where as the suggested Node 10 Buster [has 43 critical](https://hub.docker.com/layers/library/node/10-buster/images/sha256-3499a79b700369622b0b25947000d905c8ac1b67ea6f79b2f4bf1e56cf61e6f1?context=explore).  This may be a bit of a strawman as often really old releases don't have CVEs filed against them so the look more secure, but they're not
* Would Node 10 Buster Slim (`node:10-buster-slim`) be viable?  I did a quick smoke test (`docker build` and then `docker run` but w/ no postgres testing) and it didn't error out.  Aside from being ~66% smaller, it also only [has 12 critical vulns](https://hub.docker.com/layers/library/node/10-buster-slim/images/sha256-8172f6a02bac05409909ceceffae6561da693624b0d43b54050a58854e099920?context=explore) :shrug:

Issue #134, opened 2023-06-14
Support for only syncing meta databases


kennsippell:
For projects like ug-vht we have a user account with access to the `medic-users-meta` database but not the `medic` database. We aim to sync the project's metadata for monitoring but not PHI. Currently there is no option to opt-out of syncing the `medic` database.

kennsippell:
cht-couch2pg 3.6.0 has three new environment variables which control the couch databases getting synced: `SYNC_DB_MEDIC`, `SYNC_DB_SENTINEL`, and `SYNC_DB_USER_META`. Users should set these values to `false` to disable syncing the associated database

PR #135, opened 2023-06-14
feat(#134) - allow control for databases to sync

kennsippell:
#134


kennsippell:
@njogz Would you be the right reviewer for this? Unsure if couch2pg changes like this would fall under #product-ecosystems. Kindly let me know if I should work with somebody else for a review.

Issue #137, opened 2023-06-22
Sync users database to Postgres


kennsippell:
A common failure in CHT is that the roles in the `users` database do not match the `user-settings` database. We'd like to write assertions to detect these cases via cht-app-monitoring-data-ingest and request the user document in Postgres. https://github.com/medic/cht-app-monitoring-data-ingest/issues/64

garethbowen:
I'm a little concerned about the security implications of replicating the encrypted passwords. It's not dire, but still less than ideal. Is there another way to expose/detect these errors? Is there a way to have CHT automatically fix this issue rather than being a manual step? How does this error present?

kennsippell:
Discussed the security implications of this today with @mrjones-plip. I'll defer to him re security talk.

We detect these issues today through [scripts like this](https://github.com/medic/cht-pih-malawi-app/blob/main/scripts/detect-user-oddities.js). They are run manually. We are trying to integrate this with monitoring efforts like watchdog.

My understanding is that some parts of CHT use the _users database to determine a user's role and permissions (CouchDB/API/etc), while others use the user-settings doc in medic to determine the user's roles and permissions (webapp). These fall out of sync quite often (eg. I saw it twice last week [example](https://github.com/medic/config-muso/issues/987#issuecomment-1587832887)).  

Also the facility_id can be out of sync between the two.

I don't have a clear summary of how this manifests when roles are not aligned -- but I'd summarize it as "likely general spookiness".

garethbowen:
Would you mind raising an issue in cht-core for the bug (as well as continuing with monitoring)? Essentially cht-core should consistently use the one in user-settings because that's the only one that's downloaded for offline use. However I suspect couchdb db level permissions will use the one in the _users db because that's the only one it knows about. The CHT has code to keep these in sync but obviously it's not working correctly.

kennsippell:
I've opened https://github.com/medic/cht-core/issues/8337.

> as well as continuing with monitoring

Can you clarify what you mean here? Are you saying you have no blocking security concerns here and I should proceed to sync _users db to Postgres? Or are you saying we should continue to manually monitor this + address a fix for it in core?

mrjones-plip:
>> I'm a little concerned about the security implications of replicating the encrypted passwords. It's not dire, but still less than ideal.
>
> Discussed the security implications of this today with @mrjones-plip. I'll defer to him re security talk.

[...]

PR #138, opened 2023-06-23
Replicate _users database to table couchdb_medic_users

kennsippell:
#137 

Change to `couch2pg@0.7.1` is https://github.com/medic/couch2pg/pull/45


package.json
```diff
@@ -17,13 +17,13 @@
   "dependencies": {
     "chalk": "2.4.1",
     "clui": "0.3.6",
-    "couch2pg": "^0.7.0",
+    "couch2pg": "^0.7.1",
```
kennsippell, line 20: Published result of https://github.com/medic/couch2pg/pull/45

package.json
```diff
@@ -17,13 +17,13 @@
...
     "figlet": "1.2.0",
     "inquirer": "^3.0.6",
     "loglevel": "^1.4.1",
     "pg": "^8.0.0",
     "pg-format": "1.0.4",
-    "pg-promise": "^10.0.0",
+    "pg-promise": "^10.11.1",
```
kennsippell, line 26: Running `npm i` bumped `pg-promise` to 10.15.x which dropped support for NodeJS<12. This causes CI tests to fail on Node 8 and 10. I've proposed to around around the issue by pinning the version to the current production version.

package.json
```diff
@@ -17,10 +17,11 @@
   "dependencies": {
     "chalk": "2.4.1",
     "clui": "0.3.6",
-    "couch2pg": "^0.7.0",
+    "couch2pg": "^0.7.1",
     "figlet": "1.2.0",
     "inquirer": "^3.0.6",
     "loglevel": "^1.4.1",
+    "loglevel-message-prefix": "^3.0.0",
```
kennsippell, line 24: This is used directly in `libs/medic/log.js` so we should include it as a dependency.

kennsippell:
@njogz You might think you've seen this exact PR because of #136. Truly very similar! 
I appreciate your eyes here again.

env.js
```diff
@@ -23,5 +23,6 @@ module.exports = function() {
     syncSentinelDb: notFalse(process.env.SYNC_DB_SENTINEL),
     syncUserMetaDb: notFalse(process.env.SYNC_DB_USER_META),
     syncLogsDb: notFalse(process.env.SYNC_DB_LOGS),
+    syncUsersDb: process.env.SYNC_DB_USERS === 'true',
```
kennsippell, line 26: Should this still default to off now that security information is removed from the docs?

mrjones-plip: @kennsippell  - now that we cleanse the hashes before inserting to postgres, I'm OK if we default to being on.  I definitely can see the value of having user data side by side with the rest of the lot.  

Thanks for checking!

mrjones-plip: @kennsippell  - now that we cleanse the hashes before inserting to postgres, I'm OK if we default to being on.  I definitely can see the value of having user data side by side with the rest of the lot.  

Thanks for checking!

yrimal:
Are there any materialized views that you think are useful to have across all deployments? For example, we create multiple views and materialized views during initial setup based on which file is being run [here](https://github.com/medic/cht-couch2pg/tree/main/libs/analytics/migrations). It will be nice to have those views/materialized views created and executed as a part of c2pg rather than requiring us to recreate same objects later.

kennsippell:
Thanks for the feedback @yrimal. I personally find the _user data to be quite simple. The table is usually small (<12000 users max for CHT instances). We might consider a reusable view, but I'm not sure it adds that much value. Would you use it and prefer to have it?

libs/medic/runner.js
```diff
@@ -35,6 +35,11 @@ const replicateAll = async (couchUrl, pgconn, opts) => {
     if (opts.syncLogsDb) {
       await replicateDatabase(`${couchUrl}-logs`, 'couchdb_medic_logs');
     }
+
+    if (opts.syncUsersDb) {
+      const couchUrlWithoutPath = couchUrl.substring(0, couchUrl.lastIndexOf('/'));
```
dianabarsan, line 40: I dislike all url manipulation in this file. I think the URL never even gets normalized. Not something that needs to be done in this PR though.

dianabarsan (commented):
Just nits inline. Thanks for adding this!

dianabarsan (approved):

Issue #141, opened 2023-06-30
Automated releasing of cht-couch2pg


kennsippell:
v3.6.2 was deployed without github tags. Let's automate deployment of cht-couch2pg like cht-conf and other cht repositories.

kennsippell:
[Example PR from cht-conf](https://github.com/medic/cht-conf/commit/244203e341fe46e11dcb7249fd3a77dcb8d1f43c)

yrimal:
Hi @kennsippell , I'm stuck running `npm run semantic-release`

```
[semantic-release]: node version >=14.17 is required. Found v12.22.12.
```

cht-couch2pg is not supported on Node 14 or later, so I guess we'd be blocked until then.

kennsippell:
Unsure... Does semantic-release actually invoke code from within the cht-couch2pg project. If not, then I don't totally get why we can't have cht-couch2pg run on node12 while semantic-release runs separately in a different place on node14? Does [this action](https://github.com/medic/cht-couch2pg/pull/142/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R18) using Node14 not work? I don't see logs for it.

PR #142, opened 2023-07-05
chore(#141) Automate release of cht-couch2pg

1yuv:
Work for #141

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
+jobs:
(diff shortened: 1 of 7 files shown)


mrjones-plip (approved):
I'm not that close to this code but given A) it's a near clone of [this other PR](https://github.com/medic/cht-conf/commit/244203e341fe46e11dcb7249fd3a77dcb8d1f43c) that already works and B) worst case is the release will break, seems low risk.

LGTM!

kennsippell:
Nice! How are you planning to test it end to end? Any upcoming work?

PR #145, opened 2024-01-31
fix(#144): Handle null date in telemetry doc

### LINKED ISSUE #144: Handle case when telemetry dates are null
There's an issue in CHT 4.5 where telemetry documents might miss the `year`,`month`,and `day` and because of this cht-couch2pg fails for all projects running CHT 4.5 or later until medic/cht-core#8837 is fixed. 

```
[2024-01-31T19:32:18.890Z ERROR]:  error: invalid input syntax for type date: "1"
    at Parser.parseErrorMessage (/var/adapter/shared/cht-couch2pg-v3.6.2/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/var/adapter/shared/cht-couch2pg-v3.6.2/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/var/adapter/shared/cht-couch2pg-v3.6.2/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/var/adapter/shared/cht-couch2pg-v3.6.2/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (events.js:314:20)
    at addChunk (_stream_readable.js:297:12)
    at readableAddChunk (_stream_readable.js:272:9)
    at Socket.Readable.push (_stream_readable.js:213:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
  length: 235,
  severity: 'ERROR',
  code: '22007',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'SQL statement "REFRESH MATERIALIZED VIEW CONCURRENTLY useview_telemetry"\n' +
    'PL/pgSQL function refresh_matviews() line 15 at EXECUTE',
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'datetime.c',
  line: '3909',
  routine: 'DateTimeParseError',
  query: 'SELECT refresh_matviews()',

```

1yuv:
Fix #144 by assigning default date of `1970-1-1` where dates are not available.


libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql
```diff
@@ -24,26 +24,29 @@ CREATE INDEX idx_useview_feedback_period_start_user ON useview_feedback(period_s
...
-    CASE                                    --> day of the month, else 1
-      WHEN
-        (doc#>>'{metadata,day}') IS NOT NULL
-      THEN
-        doc#>>'{metadata,day}'
-      ELSE
-        '1'
-    END
-  )::date AS period_start,
+  (CASE
```
kennsippell, line 27: i don't think migrations run twice, so you'll have to create new which drop and replace this view instead of editing this definition

libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql
```diff
@@ -24,26 +24,29 @@ CREATE INDEX idx_useview_feedback_period_start_user ON useview_feedback(period_s
...
+      doc#>>'{metadata,year}',                --> year
+      CASE                                    --> month of the year
+        WHEN
+          string_to_array(substring(doc#>>'{metadata,versions,app}' FROM '(\d+.\d+.\d+)'),'.')::int[] < '{3,8,0}'::int[]
+        THEN
+          (doc#>>'{metadata,month}')::int+1   --> Legacy, months zero-indexed (0 - 11)
+        ELSE
+          (doc#>>'{metadata,month}')::int     --> Month is between 1 - 12
+      END,
+      CASE                                    --> day of the month, else 1
```
kennsippell, line 41: i know this isn't your code, but cleaner to use COALESCE here instead of CASE

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
```diff
@@ -9,26 +9,28 @@ SELECT
...
-      ELSE (doc #>> '{metadata,month}')::integer
-    END,
-    CASE
-      WHEN doc #>> '{metadata,day}' IS NOT NULL 
-      THEN doc #>> '{metadata,day}'
-      ELSE '1'
-    END
-  )::date AS period_start,
+  (CASE 
+      WHEN doc#>>'{metadata,year}' IS NULL THEN '1970-1-1'
```
kennsippell, line 13: can this be null instead of random date?

kennsippell (changes requested):
Thanks for contribution. period_start getting even wilder....

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
1yuv: Hi @kennsippell , There'll be difference in results when using `NULL` vs `1970-1-1`. Since we have done `DISTINCT ON (doc #>> '{metadata,deviceId}', doc #>> '{metadata,user}')`, If we put NULL on our field, the NON-NULL value will be selected by default which in our case is first record and old record. If we put `1970-1-1`, it will consider the latest data aailable for those user and device. Since in our case, most of the records with `null` are recent, we will not be creating metadata for recent numbers. 

With this, there'll be two cases. 

1. Keep NULL, and get older records, most likely prior to 4.5.0 upgrades, period_start will have good older date and old data will be available. 
2. Keep 1970-1-1 and get recent record for that user and device. `period_start` will be 1970-1-1 in this case and latest data available. 

Let me know what'd you prefer here.

kennsippell: IC. The `distinct on` should pick the first value ordered by `device_id, user_name, period_start`.  NULL is always sorted last by postgres. `1970` is likely also always going to be last, so they seem pretty close in functionality (?)

A third option is that we could use the 4.5.0 release date, since we know the telemetry doc is "at least that old" ? Nov-20-2023. Maybe that is getting too weird.

I think all three are decent options. You probably know best since you're deeper in the data atm. I'd likely go with NULL just to avoid "magic values" - but have no strong opinion and would respect either choice.

libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql
1yuv: I moved this code to a function so that it's more readable and single place to fix if any issue.

libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql
1yuv: Thanks, I've made the changes. 

I had to use `drop cascade` since materialized views can not be replaced like view `CREATE OR REPLACE`.

libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql
1yuv: Since we already have issue on cht-core to fix this problem of null telemetry dates, it doesn't really matter once that's fixed on what we've put here for useview_telemetry_devices. For other views, I opted to go with `NULL`.

libs/medic-users-meta/migrations/202402183030.do.telemetry_metrics.sql
```diff
@@ -0,0 +1,37 @@
+DROP MATERIALIZED VIEW IF EXISTS useview_telemetry_metrics CASCADE;
```
kennsippell: this cascade delete can cause some real problems... we did this before and the problem is that you never know what you're going to cascade drop. who knows what people have built on these?

is it possible to alter?

libs/medic-users-meta/migrations/202105171933.do.86.usersMeta.sql
kennsippell: that's pretty scary then... who knows what we are going to cascade delete... 
i'm not sure how we message this - but i think we need to be really clear about this somehow. can you ask on #development or something?

1yuv:
Closing this Pull request without merging since the issue was with Nepali projects only and we wanted to avoid drop cascade. Any projects having this issue can apply changes on this PR.

PR #153, opened 2026-07-23
feat(#152): optimize useview_telemetry_metrics

1yuv:
- Optimize useview_telemetry_metrics #152 
- Add function to handle all type of null dates #144


libs/medic-users-meta/migrations/202607231200.do.get_telemetry_date.sql
```diff
@@ -0,0 +1,23 @@
...
+    ELSE concat_ws('-',
+           year,
+           CASE
+             WHEN COALESCE(day, '') = ''
+                  AND (COALESCE(app_version, '') = ''
+                       OR string_to_array(substring(app_version, '(\d+.\d+.\d+)'), '.')::integer[] < '{3,8,0}'::integer[])
+             THEN (month::int + 1)::text
+             ELSE month
+           END,
+           COALESCE(day, '1')
```
binokaryg: Can the day be an empty string '', resulting in `COALESCE('', '1');` ?

binokaryg (approved):

libs/medic-users-meta/migrations/202607231200.do.get_telemetry_date.sql
1yuv: Good one, we haven't had the issue with this so far, but i will add this check.

PR #162, opened 2026-07-28
chore(#161): update release process to use oidc

1yuv:
#161. 

- Also updated the node version for release package, checkout version.
- Brought the changelog.md back that I removed on #160


jkuester (approved):
LGTM

PR #165, opened 2026-07-28
chore(#164): update semantic packages

1yuv:
Closes #164

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
+      contents: write # to be able to publish a github release
+      issues: write # to comment on released issues
+      pull-requests: write # to comment on released pull requests
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+lockfile-version=1
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,43 +4,60 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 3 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -9,6 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
+      contents: write # to be able to publish a github release
+      issues: write # to comment on released issues
+      pull-requests: write # to comment on released pull requests
```
jkuester, line 14: To the best of my knowledge, these should not be necessary for your current .releaserc.json.  The [eslint-config](https://github.com/medic/eslint-config/blob/master/.github/workflows/release.yml) runs with very similar config and does not need them.

.npmrc
jkuester, line 1: Is this necessary?  I don't think we set this config in other repos?

.github/workflows/release.yml
jkuester: I do see you are a major version _ahead_ of esling-config for `@semantic-release/changelog` and `@semantic-release/git`.  Perhaps this is the difference?  🤔

jkuester (approved):
Approving to unblock, though I have some questions inline.

.github/workflows/release.yml
1yuv: I was following the semantic documentation [here](https://semantic-release.org/recipes/ci-configurations/github-actions/). May be in later versions they made it required.

.npmrc
1yuv: cht-couch2pg has [issue](https://github.com/medic/cht-couch2pg/issues/82) that it does not run in later version. This prevents any npm or npm ci writing pacage-lock in later version and to be committed in not supported lock version.

